PR #3, opened 2015-08-20
Support decoding of Base64 with intermediary padding.

codematters:
Some base64 encoders in the wild incorrectly pad data before the end of
data. For instance, the string "testtesttest" might be encoded as
"dGVzdA==dGVzdA==dGVzdA==" instead of "dGVzdHRlc3R0ZXN0". This is,
technically speaking, a bug.

This patch makes the decoder tolerant to this kind of nonsense.
Intermediary padding is processed and then decoding carries on as
though a new string were provided.

### DIFF
--- a/src/main/java/org/apache/commons/codec/binary/Base64.java
+++ b/src/main/java/org/apache/commons/codec/binary/Base64.java
@@ -19,6 +19,9 @@ package org.apache.commons.codec.binary;
 
 import java.math.BigInteger;
 
+import org.apache.commons.codec.binary.BaseNCodec.Context;
+
+
 /**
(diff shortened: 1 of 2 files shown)


garydgregory:
I would be OK with a workaround like this if it had to be turned with a strict=false or lenient=true option.

kinow:
+1 to what @garydgregory suggested. I'd prefer to keep it strict by default, and have an option to make it accept this other format. Or maybe a different Base64 decoder (we currently support only RFC2045's Base64 format). And there are conflicts in the PR, so would be nice if it could be updated to fix/merge the conflicts too :-)

PR #5, opened 2016-09-13
Do not leak file descriptors

SvetlinZarev:
Bug: CODEC-225

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/DaitchMokotoffSoundex.java
+++ b/src/main/java/org/apache/commons/codec/language/DaitchMokotoffSoundex.java
@@ -232,8 +232,11 @@ public class DaitchMokotoffSoundex implements StringEncoder {
         }
 
         final Scanner scanner = new Scanner(rulesIS, CharEncoding.UTF_8);
-        parseRules(scanner, RESOURCE_FILE, RULES, FOLDINGS);
-        scanner.close();
+        try {
+            parseRules(scanner, RESOURCE_FILE, RULES, FOLDINGS);
+        } finally {
+            scanner.close();
+        }
 
         // sort RULES by pattern length in descending order
         for (final Map.Entry<Character, List<Rule>> rule : RULES.entrySet()) {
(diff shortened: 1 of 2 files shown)


jochenw:
Questions on your patch:

a) In Rule.java, you are treating the Scanner as the leakable Resource. In DaitchMokotoffSoundex.java, however, its the underlying InputStream. Sounds odd to me.
b) Use of java.util.logging: Is that standard behaviour in Codec? If not, it ought to be removed.

jochenw:
Rethinking the logger stuff: Do *not* catch the exception, or rethrow it. In general, Codec should leave the error handling to the user.

SvetlinZarev:
> In Rule.java, you are treating the Scanner as the leakable Resource. In DaitchMokotoffSoundex.java, however, its the underlying InputStream. Sounds odd to me.

In Rule I don't have access to the underlying stream, so I close the scanner which in turn closes the stream. I can change it to close the Scanner in DaitchMokotoffSoundex as well.

>  Use of java.util.logging: Is that standard behaviour in Codec? If not, it ought to be removed.

I grepped for "log, logger, etc" and I couldn't fine any, so there is no "standard" behavior.

> Rethinking the logger stuff: Do not catch the exception, or rethrow it. In general, Codec should leave the error handling to the user.

Throwing from finally will hide the original exception, that's the reason I catch it. I also log it becaue IMO it's worse to silently drop it. An alternative approach would be to close the scanner instead of the stream. The scanner.close() does not throw exception and hence no need for logging and catching it. 

What do you think ?

Svetlin

jochenw:
> In Rule I don't have access to the underlying stream, so I close the scanner which in turn closes the
>  stream. I can change it to close the Scanner in DaitchMokotoffSoundex as well.

Please, do.

Regarding the logging: I agree, that the original exception should not be hidden, which is why my code typically looks like this:

```
  Resource res = null;
  Throwable th = null;
  try {
    res = createResource();
    res.close();
    res = null;
  } catch (Throwable t) {
    th = t;
  } finally {
    if (res != null) {
      try {
        res.close();
      } catch (Throwable t) {
        if (th == null) { th = t; }
      }
   }
   if (th != null) {
     rethrow(t);
   }

```

Slightly more convoluted, but safe.

SvetlinZarev:
Made the change to close() the Scanner instead of the stream.

SvetlinZarev:
merged

PR #7, opened 2017-07-04
CODEC-224 Add methods to DigestUtils to generate hex from MessageDigests et al. 

sesuncedu:
This PR adds methods to DigestUtils to make it easy to generate hexadecimal strings from: 
- MessageDigest, 
- DigestInputStream,
- DigestOutputStream.

  If the underlying digest supports ```clone()```,  a clone is created before calling ```MessageDigest::digest``` to avoid resetting the state.  If this cannot be done, the original digest is used.

[Although clone is optional for MessageDigests, it seems to be almost universally supported (e.g. both the Sun and the IBM  providers implement clone for their concrete digests) ]

Signed-off-by: Simon Spero <sesuncedu@gmail.com>


sesuncedu:
Closing this PR because of issue with using the wrong issue number

PR #8, opened 2017-07-05
Add methods to DigestUtils to generate hex from MessageDigest et. al. 

sesuncedu:
*This PR replaces #7 which was associated with the wrong issue number*
This PR adds methods to DigestUtils to make it easy to generate hexadecimal strings from: 
- MessageDigest, 
- DigestInputStream,
- DigestOutputStream.

  If the underlying digest supports ```clone()```,  a clone is created before calling ```MessageDigest::digest``` to avoid resetting the state.  If this cannot be done, the original digest is used.

[Although clone is optional for MessageDigests, it seems to be almost universally supported (e.g. both the Sun and the IBM  providers implement clone for their concrete digests) ]

Signed-off-by: Simon Spero <sesuncedu@gmail.com>

### DIFF
--- a/src/main/java/org/apache/commons/codec/digest/DigestUtils.java
+++ b/src/main/java/org/apache/commons/codec/digest/DigestUtils.java
@@ -23,6 +23,8 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
+import java.security.DigestInputStream;
+import java.security.DigestOutputStream;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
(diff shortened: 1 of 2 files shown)


sesuncedu:
WILCO.

I will assume
1. the intended granularity of "new API" is one test method for each added
API  method and;
2. that it if there is an assertion failure  in an API method that is used
by another API method,  it is possibly but not necessarily the case that a
corresponding assertion failure be signaled in the test for the calling
method. (ie redundant, duplicate checks are OK,  but the test of the
dependent method may error due to exceptions, etc.

Also, did you know: there are a lot of methods in digestutils :)

garydgregory:
Go for it. 

Lots of methods? You should see StringUtils in Apache Commons Lang! ;-)

PR #9, opened 2017-10-02
CODEC-240 Percent-Encoding codec proposal

sermojohn:


src/main/java/org/apache/commons/codec/net/PercentCodec.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Implements the Percent-Encoding scheme, as described in HTTP 1.1 specification.
+ *
+ * @see <a href="https://tools.ietf.org/html/rfc3986#section-2.1">Percent-Encoding</a>
+ */
+public class PercentCodec implements BinaryEncoder, BinaryDecoder {
+
+    private final byte ESCAPE_CHAR = '%';
+
+    private BitSet alwaysEncodeChars = new BitSet();
```
garydgregory: Why not make this final?

src/main/java/org/apache/commons/codec/net/PercentCodec.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public PercentCodec(char[] alwaysEncodeCharArray) {
+        super();
+        fillBitSet(alwaysEncodeCharArray);
+    }
+
+    private void fillBitSet(char[] alwaysEncodeCharArray) {
+        if (alwaysEncodeCharArray == null)
+            return;
```
garydgregory: Missing { and }.

src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
```diff
@@ -0,0 +1,70 @@
...
+        PercentCodec percentCodec = new PercentCodec();
+        percentCodec.encode("test");
+    }
+
+    @Test(expected = DecoderException.class)
+    public void testDecodeUnsupportedObject() throws Exception {
+        PercentCodec percentCodec = new PercentCodec();
+        percentCodec.decode("test");
+    }
+}
```
garydgregory: It would be nice to add a code coverage report to the component; not required for this patch but a nice-to-have.

sermojohn:
I refactored the solution a bit, in order to improve the performance and support the space character to plus conversion as well. So feature-wise PercentCodec and PercentEscaper (from google-http-client) are equal. Except that I do not provide a fixed set of characters for the different URL encoding scenarios.

Regarding performace, I benchmarked both solutions using JMH beanchmark library and it seems that my solution is a bit faster (10-20% depending on the input). 

I had to do some optimizations while comparing with PercentEscaper performance, but I did not copy the byte indexing method, so I keep using BitSet for storing the special characters to always encode instead of using byte[], which as far as I can see will cause high memory utilization.

PR #14, opened 2018-09-22
[CODEC-250] correcting handling of special character between equal letters for cologne phonetic

DerXHelp:
Fix for [CODEC-250](https://issues.apache.org/jira/browse/CODEC-250) for handling of special characters between equal letters for cologne phonetic.
"Test-test" and "testtest" should result in the same value according to cologne phonetic.


DerXHelp:
PR no longer required, since it has been fixed in a different way. See CODEC-250 for more details.

PR #18, opened 2019-02-16
Fix javadoc (and Travis build)

kinow:
Should fix travis build. Removes duplicated `@param`, and adds `@return` value.


kinow:
Ok, so even though this fixes javadocs warnings, there's still an issue with modules and javadocs (I believe related to this one https://bugs.openjdk.java.net/browse/JDK-8212233). As we have not moved to Java 8 yet, better to leave the broken builds as is and avoid messing with javadocs/pom settings for now.

Will merge it unless there are objections in the next days. Happy to take any reviews in the meantime :)

PR #19, opened 2019-05-03
CODEC-134: Update commons-codec to reject decoding any impossible string encoding for Base32 and Base64.

tmousaw-ptc:
Fix CODEC-134.

### DIFF
--- a/src/test/java/org/apache/commons/codec/binary/Base64TestData.java
+++ b/src/test/java/org/apache/commons/codec/binary/Base64TestData.java
@@ -30,7 +30,7 @@ import java.util.Random;
  */
 public class Base64TestData {
 
-    public static final String CODEC_101_MULTIPLE_OF_3 = "123";
+    public static final String CODEC_101_MULTIPLE_OF_3 = "124";
 
     public static final String CODEC_98_NPE
         = "YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXpBQkNERUZHSElKS0xNTk9QUVJTVFVWV1hZWjAxMjM";
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -45,6 +45,7 @@ The <action> type attribute can be add,update,fix,remove.
 
     <release version="1.13" date="YYYY-MM-DD" description="TBD">
       <action issue="CODEC-257" dev="ggregory" type="update">Update from Java 7 to Java 8</action>      
+      <action issue="CODEC-134" dev="tmousaw-ptc" type="fix">Reject any decode request for a value that is impossible to encode to for Base32/Base64 rather than blindly decoding.</action>      
     </release>
 
     <release version="1.12" date="2019-02-04" description="Feature and fix release.">
--- a/src/main/java/org/apache/commons/codec/binary/Base32.java
+++ b/src/main/java/org/apache/commons/codec/binary/Base32.java
@@ -332,7 +332,7 @@ public class Base32 extends BaseNCodec {
      * @param inPos
      *            Position to start reading data from.
      * @param inAvail
-     *            Amount of bytes available from input for encoding.
+     *            Amount of bytes available from input for decoding.
      * @param context the context to be used
      *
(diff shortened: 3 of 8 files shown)


garydgregory:
I am not sure I like the indirection of a method call just to shift some bits. That seems over the top to me.

tmousaw-ptc:
@garydgregory - I completely agree with you. I'll rework that.

aherbert:
You are using `context.lbitWorkArea >>` through the code. I assume this is OK and the work area will never be signed but does the unsigned operator work as well? 

`context.lbitWorkArea >>>`

This makes more sense to me when pushing around bits with no care for the sign bit.

tmousaw-ptc:
@aherbert - That may be a valid concern, but that code existed before this change and, therefore, I feel it should not be changed under CODEC-134.

garydgregory:
@tmousaw-ptc ,

Would you mind doing a rebase on master one last time?

Thank you,

Gary

tmousaw-ptc:
@garydgregory - Rebased and pushed.

garydgregory:
Something went wrong with your git work. The PR now contains 91 files changed. Please redo.

tmousaw-ptc:
@garydgregory Sorry about that. Missed a step in my rebase. Should be all set now.

garydgregory:
Sorry about one more thing: You should update changes.xml with your contribution.

tmousaw-ptc:
@garydgregory - No need to apologize. This is why we review, right? Updated.

src/test/java/org/apache/commons/codec/binary/Base64TestData.java
```diff
@@ -30,7 +30,7 @@
  */
 public class Base64TestData {
 
-    public static final String CODEC_101_MULTIPLE_OF_3 = "123";
+    public static final String CODEC_101_MULTIPLE_OF_3 = "124";
```
welcor, line 33: ERhm... what's going on here?

tmousaw-ptc: I can't say that I remember changing this and I'm not sure why I would have. It clearly doesn't make any sense.

welcor: I dug in and found the reason: https://issues.apache.org/jira/browse/CODEC-134?focusedCommentId=13222167&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13222167 
It should really be renamed to "three digit valid base64 value" or someting.

tmousaw-ptc: Ah! Thank you for digging this up! I knew there was some reason behind it, but since I made this change about a year and a half ago, I didn't remember why. Should've changed the constant name while I was at it to make it more clear and possibly add a comment to explain why 124 fits the bill.

aherbert: Note that this has already been merged for `1.13`. The functionality was incomplete and was fixed in CODEC-270 for release `1.14`. We have just released `1.15` that makes this functionality opt-in rather than always on. For version `1.15` trailing bit checking is off by default thus matching the behaviour from `1.12` and earlier where invalid trailing bits are discarded. You can enable it for strict decoding where invalid bits raise an exception when decoding.

welcor: Yes, I found the strangely named variable when going through the changes in 1.15 (because 1.14 broke our builds). It just struck me that 123 is indeed a multiple of 3, but 124 isn't. Later on, I found out that it probably was meant as "valid b64 of length 3". Which works for 124 but not 123. So it's a poorly named variable.

aherbert: This variable dates back to an old bug for CODEC-101. The variable should have a byte length that is a multiple of 3. The issue is that the string `123` is not a valid base 64 encoding, so it had to change for this PR. I suggest a name change to `CODEC_101_INPUT_LENGTH_IS_MULTIPLE_OF_3`.

PR #20, opened 2019-05-31
CODEC-214: Support @NotThreadSafe @Immutable Javadoc tags

melloware:


garydgregory:
It's a bit odd to add this to the POM without actually using the tags. I'd rather wait until you actually _use_ the tags to merge something like this in the code base.

melloware:
Closing PR.

PR #23, opened 2019-07-13
CODEC-259: fixed ByteBuffer issues and updated tests

Megaprog:
#CODEC-259

### DIFF
--- a/src/test/java/org/apache/commons/codec/binary/HexTest.java
+++ b/src/test/java/org/apache/commons/codec/binary/HexTest.java
@@ -227,18 +227,19 @@ public class HexTest {
 
     @Test
     public void testDecodeByteBufferEmpty() throws DecoderException {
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/codec/binary/HexTest.java
```diff
@@ -227,18 +227,19 @@ public void testDecodeByteArrayOddCharacters() {
 
     @Test
     public void testDecodeByteBufferEmpty() throws DecoderException {
-        assertTrue(Arrays.equals(new byte[0], new Hex().decode(ByteBuffer.allocate(0))));
```
garydgregory: I would prefer to add tests instead of updating each call to `allocate` with a call to `allocateDirect`. This is something that could be done either with subclasses (one for allocate, one for allocateDirect) or with parameterizing the test with lambdas.

Megaprog: I'm sorry but have no time for that

PR #25, opened 2019-07-24
Adding power support

lysannef:
Adding support for travis job on ppc64le architecture.


lysannef:
Ppc64le job is passing.

PR #26, opened 2019-09-28
Patch for CODEC-239 - Allow "look-alike" characters when decoding Base32

ChristopherSchultz:
Allows "look-alike" characters like 0 (digit zero) and 1 (digit one) as aliases for O (uppercase oh) and I (uppercase eye) respectively when decoding.

### DIFF
--- a/src/main/java/org/apache/commons/codec/binary/Base32.java
+++ b/src/main/java/org/apache/commons/codec/binary/Base32.java
@@ -77,6 +77,25 @@ public class Base32 extends BaseNCodec {
     };
 
     /**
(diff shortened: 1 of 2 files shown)


ChristopherSchultz:
Gary Gregory requested in https://issues.apache.org/jira/browse/CODEC-239 that I create a PR for this. I finally got around to doing it.

garydgregory:
Upon reexamination, I think that what this use case is looking for is Base58, see [CODEC-326](https://issues.apache.org/jira/browse/CODEC-326). I think we should close this ticket in favor of [CODEC-326](https://issues.apache.org/jira/browse/CODEC-326)

garydgregory:
Closing: No feedback.

PR #29, opened 2019-11-24
[CODEC-270] Base32/64: Fix masked check of the final bits to discard.

aherbert:
Fixed the Base32/64 mask check to ensure it checks all the bits to be discarded are zero.

Added tests that enumerate all possible trailing characters. Only those with zero for the discarded bits should be decoded.

Also fixed the BCodecTest data to use the RFC 1522 "encoded-word" header.

### DIFF
--- a/src/main/java/org/apache/commons/codec/binary/Base32.java
+++ b/src/main/java/org/apache/commons/codec/binary/Base32.java
@@ -114,8 +114,20 @@ public class Base32 extends BaseNCodec {
             'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
     };
 
+    /** Mask used to extract 7 bits, used when decoding final trailing character. */
+    private static final long MASK_7BITS = 0x7fL;
+    /** Mask used to extract 6 bits, used when decoding final trailing character. */
+    private static final long MASK_6BITS = 0x3fL;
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/commons/codec/binary/Base32.java
```diff
@@ -548,19 +561,23 @@ public boolean isInAlphabet(final byte octet) {
...
      * @throws IllegalArgumentException if the bits being checked contain any non-zero value
      */
-    private void validateCharacter(final int numBits, final Context context) {
-        if ((context.lbitWorkArea & numBits) != 0) {
-            throw new IllegalArgumentException(
-                "Last encoded character (before the paddings if any) is a valid base 32 alphabet but not a possible value");
+    private static void validateCharacter(final long emptyBitsMask, final Context context) {
+        // Use the long bit work area
+        if ((context.lbitWorkArea & emptyBitsMask) != 0) {
+        throw new IllegalArgumentException(
```
garydgregory: Indentation is not quite right for the guts of this `if` block.

aherbert: Sorry. I seem to have dropped 4 space indentation somehow. I'll fix that now.

PR #30, opened 2019-11-26
[CODEC-265] BaseNCodec to expand buffer using overflow conscious code.

aherbert:
Re: [CODEC-265] Add a test to encode a 1 Gigabyte byte[].

Added overflow conscious code for buffer expansion during encoding and decoding. This is based on java.util.ArrayList but modified to use a local copy of JDK 1.8 Integer.compareUnsigned(int, int) to compare the new buffer capacity to the required capacity.
    
Unlike the java.util.ArrayList if the minCapacity exceeds the head room for the maximum array size, the capacity returned is the required minCapacity and not Integer.MAX_VALUE. This allows the test `testEnsureBufferSizeExpandsToBeyondMaxBufferSize` to pass on my machine using:

```
java -version
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1ubuntu1~16.04.1-b10)
OpenJDK 64-Bit Server VM (build 25.222-b10, mixed mode)
``` 

If I return Integer.MAX_VALUE for the new capacity then the test fails as my machine cannot allocate arrays of this size. This test has an assume method that tries to allocate the very large raw byte array. If this is not possible then the test is skipped.

This fix allows reinstating Base64 HugeLineSeparator test as the buffer expansion ensures that the minimum allowed capacity is allocated even if doubling the buffer size is not enough.

Note:

The test for allocating the maximum buffer size requires >2GB of RAM. The test for encoding a 1 GB file requires >3GB of RAM. The tests attempt to detect the amount of RAM that can be allocated and skip if necessary. Travis does not run the test for encoding a 1GB file as it does not have 3GB of memory.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -43,6 +43,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
 
     <release version="1.14" date="TBD" description="Feature and fix release.">
+      <action issue="CODEC-265" dev="aherbert" type="fix">BaseNCodec to expand buffer using overflow conscious code.</action>
       <action issue="CODEC-270" dev="aherbert" type="fix">Base32/64: Fixed decoding check that all the final trailing bits to discard are zero.</action>
       <action issue="CODEC-264" dev="aherbert" due-to="Claude Warren" type="add">Add MurmurHash3.hash128x64 methods to fix sign extension error during seeding in hash128 methods.</action>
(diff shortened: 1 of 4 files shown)


aherbert:
As expected the coverage drops a touch. There is 1 line of code not covered in method `BaseNCodec.createPositiveCapacity` where the minimum capacity is close to 2GB. This is because Travis is skipping the test as it does not have enough memory to run it. Locally running the test shows full coverage:

```
mvn clean test -Dtest=BaseNCodecTest jacoco:report
```

See `target/site/jacoco/index.html`

PR #32, opened 2019-12-02
[codec-259] Hex: consume all ByteBuffer.remaining() bytes

aherbert:
The use of ByteBuffer.array() for any array-backed buffers means that the Hex class can produce incorrect results if the backing array is not entirely filled with the data to encode/decode.

This fixes the implementation by only using the backing array if the length is equal to the reported capacity. Otherwise the methods using ByteBuffer default to copying the remaining bytes into a copy array.

The code has been documented as using all remaining bytes in the input ByteBuffer. The methods now handle input array backed ByteBuffers and directly allocated ByteBuffers equally.

Tests have been added using limits on the buffer to encode/decode a sub-section of the bytes.

The tests have been updated so all tests are run with a direct allocated ByteBuffer in AllocateDirectHexTest.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -43,6 +43,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
 
     <release version="1.14" date="TBD" description="Feature and fix release.">
+      <action issue="CODEC-259" dev="aherbert" type="add">Hex: Only use an available ByteBuffer backing array if the length equals the remaining byte count.</action>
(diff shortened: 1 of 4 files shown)


aherbert:
Coverage drops because the HexTest no longer uses:
```
digest.StringUtils:
    public static ByteBuffer getByteBufferUtf8(final String string) {
```

This method seems to have been introduced for the sole purpose of use in the HexTest. There is no other usage in the codec code. It is out of place in digest.StringUtils as there are no other methods for other encodings that return a ByteBuffer. All the others return byte[].

I can add a test to StringUtilsTest to get coverage back. I think the method should be deprecated.

PR #33, opened 2019-12-03
Added string helper method to MurmurHash3.hash64

ken-kentan:
add string helper method to MurmurHash3.hash64.

```
public static long hash64(final String data)
```

### DIFF
--- a/src/main/java/org/apache/commons/codec/digest/MurmurHash3.java
+++ b/src/main/java/org/apache/commons/codec/digest/MurmurHash3.java
@@ -558,6 +558,27 @@ public final class MurmurHash3 {
     }
 
     /**
+     * Generates 64-bit hash from a string with a default seed.
+     * The string is converted to bytes using the default encoding.
+     * This is a helper method that will produce the same result as:
+     *
+     * <pre>
+     * int offset = 0;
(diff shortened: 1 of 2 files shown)


aherbert:
I am against this:

- It adds more convenience methods which do not save very much typing
- It promotes the hash64 method which is not part of MurmurHash3
- It uses String::getBytes without a character encoding

This method saves a few characters over the alternative:

```
String s;
long h1 = MurmurHash3.hash64(s);
long h2 = MurmurHash3.hash64(s.getBytes());
```

The hash64 methods are not part of the original MurmurHash3 implementation. We do not know the statistical properties of these methods, i.e. the hash collision rate. The methods were ported from Apache Hive but the original reason for the method and its usage is unknown.

The javadoc should be more clear that these methods do not return the same as either the upper of lower 64-bits of the hash128 method:

```
String s;
long[] h = MurmurHash3.hash128(s.getBytes());
long h2 = MurmurHash3.hash64(s.getBytes());
// This is not true only due to random hash collision:
// h[0] != h2 && h[1] != h2;
```

The main javadoc for hash64 states this but the other helper methods do not.

ken-kentan:
ok🙆

PR #34, opened 2019-12-28
Added javadoc messages concerning hash64 seed and sign extension.

Claudenw:
Fix for CODEC-275

### DIFF
--- a/src/main/java/org/apache/commons/codec/digest/MurmurHash3.java
+++ b/src/main/java/org/apache/commons/codec/digest/MurmurHash3.java
@@ -428,6 +428,9 @@ public final class MurmurHash3 {
      * {@linkplain #hash128x64(byte[])} with the same byte data from the {@code long}.
      * This method will be removed in a future release.</p>
      *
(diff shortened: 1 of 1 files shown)


aherbert:
I am 50/50 on this.

I would not call it a bug when there is no reference hash64 implementation. This method is a port of a bad adaption of the hash128x86 method to output a 64-bit hash. So unlike the other method where it was a bug because the output was not the same as the reference implementation, this can be labelled as an implementation detail.

The only real effect is that the initial hash will be xor'd with 1's instead of 0's for the upper 32-bits and if the seed is negative.

Since we do not know about the statistical properties of this hash, it is not part of the original code and it is marked as deprecated then adding a lot of javadoc about sign-extension bugs which do not manifest (because the default seed is positive) to 5 deprecated methods seems like noise to me. 

I would update the javadoc on the main method which accepts a seed to state that the seed is **not** treated as **unsigned**.

garydgregory:
I am going to bring this in because it helps understand the code more, especially since these comments are all on deprecated methods.

garydgregory:
https://issues.apache.org/jira/browse/CODEC-275

PR #35, opened 2020-01-23
[CODEC-280] Added strict decoding property to BaseNCodec.

aherbert:
The default is lenient encoding. Any trailing bits that cannot be
interpreted as part of the encoding are discarded. Combinations of
trailing characters that cannot be created by a valid encoding are
partially interpreted.

If set to strict encoding the codec will raise an exception when
trailing bits are present that are not part of a valid encoding.

Added tests to show that in lenient mode re-encoding will create a
different byte array.

In strict mode the re-encoding should create the same byte array as the
original (with appropriate pad characters on the input).

The PR does not current address behaviour of BCodec class that uses Base64. The test that validates input that requires strict mode have been disabled to allow travis to pass.

### DIFF
--- a/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
+++ b/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
@@ -231,7 +231,7 @@ public abstract class BaseNCodec implements BinaryEncoder, BinaryDecoder {
     }
 
     /**
-     * Sets the decoding behaviour when the input bytes contain leftover trailing bits that 
+     * Sets the decoding behaviour when the input bytes contain leftover trailing bits that
      * cannot be created by a valid encoding. These can be bits that are unused from the final
      * character or entire characters. The default mode is lenient decoding. Set this to
      * {@code true} to enable strict decoding.
@@ -242,7 +242,7 @@ public abstract class BaseNCodec implements BinaryEncoder, BinaryDecoder {
      *     are not part of a valid encoding. Any unused bits from the final character must
      *     be zero. Impossible counts of entire final characters are not allowed.
      * </ul>
-     * 
+     *
      * <p>When strict decoding is enabled it is expected that the decoded bytes will be re-encoded
      * to a byte array that matches the original, i.e. no changes occur on the final
      * character. This requires that the input bytes use the same padding and alphabet
--- a/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
+++ b/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
@@ -23,6 +23,7 @@ import org.apache.commons.codec.BinaryDecoder;
 import org.apache.commons.codec.BinaryEncoder;
 import org.apache.commons.codec.DecoderException;
 import org.apache.commons.codec.EncoderException;
+import org.apache.commons.codec.binary.BaseNCodec.Context;
 
 /**
  * Abstract superclass for Base-N encoders and decoders.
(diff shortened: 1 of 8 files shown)


src/test/java/org/apache/commons/codec/net/BCodecTest.java
```diff
@@ -157,6 +158,7 @@ public void testDecodeObjects() throws Exception {
     }
 
     @Test
+    @Ignore("CODEC-280: Disabled strict decoding by default. The BCodec uses the default so this test does not fail the impossible cases.")
```
garydgregory: @aherbert 
I think this test should be rewritten to match the new behavior, not `@Ignore`'d.

aherbert: The new behaviour is lenient. So that would be a behaviour change for BCodec and make this test redundant.

I think that the new strict decoding property can be added here too so users have an option.

garydgregory: I'm all for making Base32 and Base64 carry the same option.

aherbert: Hi @garydgregory 

I received an email from github with another comment from you but cannot find it through the web interface so I'll repeat for clarity the comment was "I'm all for making Base32 and Base64 carry the same option."

This PR already has the public setStrictDecoding(boolean) for Base32 and Base64. The unresolved is whether the property should be added to BCodec as well. At the moment this test is ignored because BCodec cannot be made strict. So either:

1. Drop the test and BCodec is always lenient
2. Make BCodec always strict (this test will pass)
3. Make BCodec have configurable strict/lenient decoding

My option would be number 3.

src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
```diff
@@ -223,6 +230,47 @@ protected BaseNCodec(final int unencodedBlockSize, final int encodedBlockSize,
...
+     * <p>When strict decoding is enabled it is expected that the decoded bytes will be re-encoded
+     * to a byte array that matches the original, i.e. no changes occur on the final
+     * character. This requires that the input bytes use the same padding and alphabet
+     * as the encoder.
+     *
+     * @param strictDecoding Set to true to enable strict decoding; otherwise use lenient decoding.
+     * @see #encode(byte[])
+     * @since 1.15
+     */
+    public void setStrictDecoding(boolean strictDecoding) {
```
boris-github, line 255: @aherbert looking at the [BaseNCodec](https://commons.apache.org/proper/commons-codec/archives/1.15/apidocs/org/apache/commons/codec/binary/BaseNCodec.html) Javadoc I can't find the **setStrictDecoding()** method, however **isStrictDecoding()** is available. Can you please double check?

And following [your example](https://issues.apache.org/jira/browse/CODEC-280) for `codec.setStrictDecoding(true)` I'm getting a compilation error when using [commons-codec 1.15](https://search.maven.org/artifact/commons-codec/commons-codec/1.15/jar):
```
symbol:   method setStrictDecoding(boolean)
location: variable codec of type org.apache.commons.codec.binary.Base32
```

boris-github: Oh, I see why. Browsing the history of the file, I found that it was modified in [Reimplement the new-in-1.15 BaseNCodec's and friends' strict vs. lenient](https://github.com/apache/commons-codec/commit/9f1b740a17f0d54366edfb45df0636b8e302666a#diff-2cf424122556d939f7804a29ae0c501d22e744e725c1a239efd0bf687ccd7402). Sadly the [release notes](https://commons.apache.org/proper/commons-codec/changes-report.html#a1.15) haven't been updated to reflect that.
(cost me 2 hours)

PR #36, opened 2020-02-18
CODEC-281: Double metaphone  encode kc as K

sguill:
**Double metaphone should handle 'K' for 'kc' entries instead of 'KK'**

One example is Kir***kc***aldy: which is presently encoded as "KR***KK***".

When omitting 'k' or 'c' letters (for example, Kircaldy or Kirkaldy) it is encoded as "KRKL".

The correction consists to verify when letter 'c' follows 'k', if 'c' is encoded as 'K', it is ignored as for a succession of 'kk' letters. With this correction, Kirkcaldy would be encoded as "KRKL" which is more discriminant and compliant with the phonetic.

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/DoubleMetaphone.java
+++ b/src/main/java/org/apache/commons/codec/language/DoubleMetaphone.java
@@ -132,8 +132,7 @@ public class DoubleMetaphone implements StringEncoder {
                 index = handleJ(value, result, index, slavoGermanic);
                 break;
             case 'K':
-                result.append('K');
-                index = charAt(value, index + 1) == 'K' ? index + 2 : index + 1;
+                index = handleK(value, result, index);
                 break;
             case 'L':
                 index = handleL(value, result, index);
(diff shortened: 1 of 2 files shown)


Ben-Waters:
Having a look, I found what I believe to be the[ vanilla version ](http://aspell.net/metaphone/dmetaph.cpp)of the algorithm from [it's author](http://aspell.net/metaphone/). 

```
case 'K':
        if(GetAt(current + 1) == 'K')
                current += 2;
        else
                current += 1;
        MetaphAdd("K");
        break;
```
It doesn't appear to do the additional check to handle 'c' to me. Is this an improvement outside of the official algorithm or is there a source for this change?

garydgregory:
@sguill 
Would you please reply to @Ben-Waters's question?

garydgregory:
@sguill ping?

sguill:
Yes, it is an improvement outside of the official algorithm.

garydgregory:
@sguill 
See my question in https://github.com/apache/commons-codec/pull/36#issuecomment-3725715717

sguill:
> Having a look, I found what I believe to be the[ vanilla version ](http://aspell.net/metaphone/dmetaph.cpp)of the algorithm from [its author](http://aspell.net/metaphone/).
> 
> ```
> case 'K':
>         if(GetAt(current + 1) == 'K')
>                 current += 2;
>         else
>                 current += 1;
>         MetaphAdd("K");
>         break;
> ```
> 
> It doesn't appear to do the additional check to handle 'c' to me. Is this an improvement outside of the official algorithm or is there a source for this change?

Yes, this is an improvement outside of the official algorithm, there is no source for this change.

garydgregory:
If there is a true use case for this change, I propose that we create an extension of `DoubleMetaphone` called `ApacheDoubleMetaphone` (or a better name). Currently private methods can become package-private to allow for easier extension.

PR #38, opened 2020-03-01
CODEC-284 update to hamcrest v2.2

nhojpatrick:
Handles https://issues.apache.org/jira/browse/CODEC-284

PR #39, opened 2020-03-01
CODEC-285 upgrade to JUnit v5.6.0

nhojpatrick:
Handles https://issues.apache.org/jira/browse/CODEC-285

Also based upon https://issues.apache.org/jira/browse/CODEC-284 branch so pr https://github.com/apache/commons-codec/pull/38

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,6 +255,18 @@ limitations under the License.
       <version>2.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <version>5.6.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.vintage</groupId>
+      <artifactId>junit-vintage-engine</artifactId>
+      <version>5.6.0</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
--- a/pom.xml
+++ b/pom.xml
@@ -236,11 +236,23 @@ limitations under the License.
       <artifactId>junit</artifactId>
       <version>4.13.2</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
+          <artifactId>hamcrest-core</artifactId>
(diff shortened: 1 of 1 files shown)


aherbert:
When including multiple JUnit dependencies I would use the JUnit BOM in the dependency management to ensure they are kept in sync:
```
<dependency>
  <groupId>org.junit</groupId>
  <artifactId>junit-bom</artifactId>
  <version>X.Y.Z</version>
  <scope>import</scope>
  <type>pom</type>
</dependency>
```
I'd like to highlight that codec targets JDK 7. These changes to the build tools would prevent building on JDK 7. I'm fine with forcing JDK 8+ for the toolchain but Java 7 still has support until July 2022 and someone may care about this.

garydgregory:
My preference would be for Codec to update to Java 8.

nhojpatrick:
> My preference would be for Codec to update to Java 8.

await another Jira/PR, I'm finding that for assertThrows i need lambdas and it would make it easier

nhojpatrick:
> When including multiple JUnit dependencies I would use the JUnit BOM in the dependency management to ensure they are kept in sync:
> 
> ```
> <dependency>
>   <groupId>org.junit</groupId>
>   <artifactId>junit-bom</artifactId>
>   <version>X.Y.Z</version>
>   <scope>import</scope>
>   <type>pom</type>
> </dependency>
> ```
> 
> I'd like to highlight that codec targets JDK 7. These changes to the build tools would prevent building on JDK 7. I'm fine with forcing JDK 8+ for the toolchain but Java 7 still has support until July 2022 and someone may care about this.

if i can upgrade all the tests quick enough, junit-vintage-engine will be going asap, so i would prefer to keep both separately defined so i can start moving tests from vintage to jupiter, then another future to drop junit-vintage-engine. i agree long term it would be better to define to keep in sync but depending how long the discussions are, depends how long you have duplicate definitions.

garydgregory:
Git master is now on Java 8. It's confusing to have both JUnit 4 AND 5 on the classpath. It should be one or the other, and yes, it would be nice to update the whole component to JUnit 5.

nhojpatrick:
@garydgregory It can be useful as your migrating from JUnit v4 to JUnit v5, otherwise is a bing bang. Take advantage of assertAll and assertThrows as that would work in JUnit v4 code.
Eventually all the JUnit v4 tests would be updated and you can remove that dependency.

PR #41, opened 2020-03-03
CODEC-285 Better StringEncoderAbstractTest testEncodeNull

nhojpatrick:

This test doesn't ever throw the exception listed, so why is it catching and ignoring if thrown???


nhojpatrick:
If easier I can collapse these into the CODEC-285 PR

PR #42, opened 2020-03-03
CODEC-285 Base32InputStreamTest testReadNull

nhojpatrick:

Using assertThrows to explicitly check the expected line throws the expected exception.


nhojpatrick:
If easier I can collapse these into the CODEC-285 PR

garydgregory:
Closing: Superseded in git master (commit 5a03f5a493240f26f27ccabe6b8ac7b701c2d5c5)

PR #43, opened 2020-03-04
CODEC-285 Base32InputStreamTest testReadOutOfBounds

nhojpatrick:

Using assertAll and assertThrows to explicitly check the expected line throws the expected exception, and all done in parallel not sequentially.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,6 +274,9 @@ limitations under the License.
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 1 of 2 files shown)


nhojpatrick:
Not checking exception message it like just checking a method returns a non null value.
Anyone changing a method would expect to have to fix any tests they break, so why not any tests asserting the message?
I've done this multiple times as discovered tests are not checking what your expecting, they are failing with the correct throwable type but not for the correct reason.

It's not like I'm checking the stackstrace and asserting the correct line that the test is expecting is throwing the exception.

garydgregory:
Closing: Superseded in git master (commit https://github.com/apache/commons-codec/commit/5a03f5a493240f26f27ccabe6b8ac7b701c2d5c5)

PR #44, opened 2020-03-06
Update DigestUtils.java

YYTVicky:


YYTVicky:
Hi, 
  Thank you for your feedback, we just want to highlight that hash algorithm like MD5, MD2 and SHA1 (refer the document here: https://www.securityfocus.com/bid/11849/discuss)have been proved been broken, we recommend you use SHA256 or SHA512 when to invoke Messagedigest.genInstance() function. 

Hope to hear any feedback from you!
~

src/main/java/org/apache/commons/codec/digest/DigestUtils.java
```diff
@@ -192,6 +192,7 @@ public static MessageDigest getDigest(final String algorithm) {
      */
     public static MessageDigest getDigest(final String algorithm, final MessageDigest defaultMessageDigest) {
         try {
+            /** potential insecure algorithm (MD2, MD5, SHA1, SHA256) called here */
```
garydgregory, line 195: @YYTVicky 
This would better done in the method Javadoc for the algorithm param.

YYTVicky: Hi @garydgregory , Thanks a lot for your kind reply, we are a security research team at Virginia Tech. We are doing an empirical study about the usefulness of the existing security vulnerability detection tools. May we follow up on several questions: 
 **Is the bug report helpful?**
**will you prefer to fix the reported vulnerability**
**Are there any types of bugs/security vulnerabilities you want the detection tools to pay more attention to?**
Appreciate any feedback from your side!

garydgregory: "Is the bug report helpful?"
No because it ignores documented behavior, I consider it a false positive.

garydgregory:
Closing, no further input from OP.

PR #48, opened 2020-05-26
fix travis-ci scripts

XenoAmess:


XenoAmess:
> I wonder if we need both openjdk11 and oraclejdk11 ?

I thought it must be a reason to contain both in original travis-ci scripts, so I didn't remove any of them.

.travis.yml
```diff
@@ -15,23 +15,23 @@
...
-  - openjdk8
-  - openjdk11
-  - openjdk12
-  - openjdk13
-  - openjdk14  
-  - openjdk-ea
-
+    - jdk: openjdk8
+    - jdk: openjdk11
+    - jdk: openjdk12
```
aherbert, line 27: Can you drop:
```
openjdk12
openjdk13
oraclejdk11
```

Now Java 14 is out 12 and 13 are redundant. Testing on oraclejdk11 and openjdk11 also seems redundant.

XenoAmess: > Now Java 14 is out 12 and 13 are redundant. Testing on oraclejdk11 and openjdk11 also seems redundant.

we can drop 12 but 13 but I think we'd better remain 11, as 11 is LTS.

aherbert: My point was that testing on **both** oraclejdk11 and openjdk11 seems redundant. I would also drop oraclejdk11 and just leave openjdk11.

XenoAmess: > My point was that testing on **both** oraclejdk11 and openjdk11 seems redundant. I would also drop oraclejdk11 and just leave openjdk11.

@aherbert oh, got it:)

PR #58, opened 2020-08-15
Takari maven-wrapper

nhojpatrick:
Maven 3.7.0 will introduce a maven wrapper that is based upon the takari maven wrapper.

The advantage of using a maven wrapper is that it puts the control over what maven version to be use into source control, and each branch or tag controls the version it requires. Both developers and CI/CD no longer have to play russian roulette over what version needs to be installed.

This PR changes mvn CI/CD commands to mvnw include documentation and comments.

I've been using takari for about 4 years and have tested the version which will released as part of maven v3.7.0. Moving to the apache maven version will simply be a case of properties file url which happens for any maven upgrade and the mvnw scripts mainly around the license header.


chtompki:
Will try to get to looking at this today. Many thanks for the effort here :-)

nhojpatrick:
@chtompki I'm going to be doing similar all afternoon for all the other apache commons projects i actively/regularly use

.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2007-present the original author or authors.
```
garydgregory, line 2: I'm not sure "-present" is valid but IANAL...

garydgregory:
This seems like a lot to add and _maintain_ for each Apache Component... and who is going to maintain so many copies? Esp when someone will hack one script and not another... yikes!
I understand the idea, but the implementation seems like a maintenance headache. Let's discuss this on the dev ML before moving on.

nhojpatrick:
@garydgregory the scripts should not be hacked locally. I executed `mvn -N io.takari:maven:0.7.7:wrapper` and added the changes. Maven v3.7.0 will have a similar command line once v3.7.0 is released.

From experience the mvnw and mvnw.cmd and change very little and each time it's been updated i've just done `mvn -N io.takari:maven:NEW_VERSION:wrapper` and committed the changes that plugin did.

Honestly it results in less maintenance and is easier for developers once you switch. Say commons-codec needs maven v3.6.0, well that version is just configure via this file `.mvn/wrapper/maven-wrapper.properties` using this line `distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.0/apache-maven-3.6.0-bin.zip`

Your CI/CD servers etc don't need maven pre installed
Any new developers and contributors don't need to install v3.6.0 and change their paths.

Say commons-codec now wants to upgrade to v3.6.3, change `.mvn/wrapper/maven-wrapper.properties` to contain `distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.6.3/apache-maven-3.6.3-bin.zip` 

CI/CD and Developers just keep executing ./mvnw and v3.6.3 is automatically downloaded and caches into your ~/.m2/wrappers. So no need to change your CI/CD docker images, or jenkins slaves to add v3.6.3.

.mvn/wrapper/MavenWrapperDownloader.java
nhojpatrick: This file is maintained by takari, and is what `mvn -N io.takari:maven:0.7.7:wrapper` created. When moving to the maven v3.7.0 version I believe it had more formal apache maven headers and copyright.

nhojpatrick:
@garydgregory have sent email to dev ML regarding these PR and also what maven wrapper is etc

nhojpatrick:
Withdrawing PR due to negative email voting regarding mvnw as being an improvement.

PR #64, opened 2020-10-13
Add form feed and vertical tab whitespace to isWhiteSpace()

jamesgiu:
Ran into an issue with these characters not being detected and having false positive `isBase64()` detection as a result - would be good to get them counted as whitespace.

Ref. C/C++ `isspace()` function considers these characters as whitespace.
http://www.cplusplus.com/reference/cctype/isspace/
https://www.oreilly.com/library/view/c-in-a/0596006977/re129.html

Cheers!

### DIFF
--- a/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
+++ b/src/main/java/org/apache/commons/codec/binary/BaseNCodec.java
@@ -261,6 +261,9 @@ public abstract class BaseNCodec implements BinaryEncoder, BinaryDecoder {
             case '\n' :
             case '\r' :
             case '\t' :
+            case '\f' :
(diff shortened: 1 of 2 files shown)


jamesgiu:
Hi all, any chance this will be looked at? Cheers :)

garydgregory:
This method or PR is not great because we are defining our own whitespace instead of reusing Charater.isWhitespace or Charater.isSpaceChar. 

Shouldn't we be using Charater.isSomething?

garydgregory:
Closing: In git master now: Deprecate `BaseNCodec.isWhiteSpace(byte)` and use `Character.isWhitespace(int)`

PR #66, opened 2020-11-25
(doc) Remove duplicate "from"s in javadoc comments

jamesylgan:
Noticed a minor javadocs issue - some classes had "from from" instead of "from"

Quick PR to fix this. Does not impact code so should not need any extra testing.

(I noticed I messed up adding "(doc)" to the actual commit - I can submit a revised PR or force push an updated commit message if needed - it also seems like I shouldn't need to sign a CLA due to how minor this is, but let me know if there's anything I'm missing :) )


aherbert:
Thanks for the corrections.

PR #79, opened 2021-03-07
CODEC-296: Add support for Blake3 family of hashes

jvz:
https://issues.apache.org/jira/browse/CODEC-296

Note that I've ported this from my own library https://github.com/o1c-dev/o1c which is ISC-licensed and based on [public domain code](https://github.com/BLAKE3-team/BLAKE3/blob/master/src/lib.rs).


jvz:
By the way, I just want to make sure I'm following the appropriate style guides and such. I don't really expect an algorithmic review here since it's an externally defined one.

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,455 @@
...
+        engineState.inputData(in, offset, length);
+    }
+
+    /**
+     * Squeezes hash output data that depends on the sequence of absorbed bytes preceding this invocation and any previously
+     * squeezed bytes.
+     *
+     * @param out destination array to squeeze bytes into
+     */
+    public void squeeze(final byte[] out) {
```
garydgregory, line 97: I do not think a lot of these methods need to be public. I think most uses cases would be "hash this" without concern about the details. The less public APIs footprint we present, the more freedom with have to change the class.

src/test/java/org/apache/commons/codec/digest/Blake3TestVectorsTest.java
```diff
@@ -0,0 +1,314 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertArrayEquals;
+
+// Each test is an input length and three outputs, one for each of the hash, keyed_hash, and derive_key modes.
```
garydgregory, line 30: Make this a Javadoc.

src/test/java/org/apache/commons/codec/digest/Blake3TestVectorsTest.java
```diff
@@ -0,0 +1,314 @@
...
+    private final Blake3 hasher = Blake3.initHash();
+    private final Blake3 keyedHasher = Blake3.initKeyedHash(KEY);
+    private final Blake3 kdfHasher = Blake3.initKeyDerivationFunction(CTX);
+
+    private final byte[] input;
+    private final byte[] hash;
+    private final byte[] keyedHash;
+    private final byte[] deriveKey;
+
+    public Blake3TestVectorsTest(int inputLength, String hash, String keyedHash, String deriveKey)
```
garydgregory, line 267: It would be nice to add a simple Javadoc and some input validation.

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,455 @@
...
+            final int outputFlags = flags | startFlag() | CHUNK_END;
+            return new Output(chainingValue, blockWords, chunkCounter, blockLength, outputFlags);
+        }
+    }
+
+    private static class EngineState {
+        private final int[] key;
+        private final int flags;
+        // Space for 54 subtree chaining values: 2^54 * CHUNK_LEN = 2^64
+        private final int[][] cvStack = new int[54][];
```
garydgregory, line 379: `54` looks magical, add a Javadoc'd constant IMO.

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,455 @@
...
+                final int take = Math.min(want, length);
+                System.arraycopy(input, offset, block, blockLength, take);
+                blockLength += take;
+                offset += take;
+                length -= take;
+            }
+        }
+
+        Output output() {
+            final int[] blockWords = unpackInts(block, 16);
```
garydgregory, line 369: `16` looks magical, add a Javadoc'd constant IMO. Does it have to match the `16` above in `update(final byte[] input, int offset, int length)`? Using a constant would avoid the question ;-)

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,455 @@
...
+public final class Blake3 {
+    // TODO: migrate to Integer.BYTES after upgrading to Java 8
+    private static final int INT_BYTES = Integer.SIZE / Byte.SIZE;
+
+    private static final int BLOCK_LEN = 64;
+    private static final int KEY_LEN = 32;
+    private static final int OUT_LEN = 32;
+    private static final int CHUNK_LEN = 1024;
+
+    private static final int[] IV =
```
garydgregory, line 44: Needs Javadoc, too magical ;-)

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,455 @@
...
+ * and a default output length of 256 bits (32 bytes) which can extended up to 2<sup>64</sup> bytes.
+ * <p>
+ * Adapted from the ISC-licensed O(1) Cryptography library by Matt Sicker and ported from the reference public domain
+ * implementation by Jack O'Connor.
+ * </p>
+ *
+ * @see <a href="https://github.com/BLAKE3-team/BLAKE3">BLAKE3 hash function</a>
+ * @since 1.16
+ */
+public final class Blake3 {
```
garydgregory, line 35: A hash "Hello World!" code example in the Javadoc would be nice.

garydgregory (changes requested):
Hi @jvz 
Thank you for the PR. I've sprinkled some comments throughout but not about the actual implementation... :-)

src/main/java/org/apache/commons/codec/digest/Blake3.java
jvz: The original name for this method is `finalize` which doesn't mesh too well with Java (similar to why the crypto APIs use `doFinal()` instead of `final()`). Perhaps I could use `doFinalize()`? This method is used for calculating the hash and can be done incrementally (e.g., knowing the hash of a part of a file as it's streaming) and being an extensible output function, this can "squeeze" as many bytes as requested to create an arbitrarily long hash (up to 2^64 bytes I think).

The squeeze/absorb naming is something I borrowed from internals of SHA-3, so I'm certainly up for renaming them to something else. Absorb is equivalent to `update` in `MessageDigest` and `Mac`.

src/main/java/org/apache/commons/codec/digest/Blake3.java
jvz: Yes, some better javadocs about the space limitations. As hinted by the vague comment left here (and explained in more detail in the linked PDF for blake3), a stack of 54 int arrays (each being 8 ints) is the maximum amount of hashing state that can accumulate when hashing the largest amount of data allowed by the algorithm. When used in applications with a smaller processing limit (e.g., some embedded device only processing kB-order messages), the stack space can be reduced. This could, of course, use something like ArrayDeque here, though I opted to avoid generating garbage or other wrappers where possible while porting the rust code.

src/main/java/org/apache/commons/codec/digest/Blake3.java
```diff
@@ -0,0 +1,505 @@
...
+        for (int i = 0; i < state.length / 2; i++) {
+            state[i] ^= state[i + 8];
+            state[i + 8] ^= chainingValue[i];
+        }
+        return state;
+    }
+
+    private static Output parentOutput(
+            final int[] leftChildCV, final int[] rightChildCV, final int[] key, final int flags) {
+        final int[] blockWords = Arrays.copyOf(leftChildCV, BLOCK_INTS);
```
jvz, line 315: Not sure if it's worth commenting in the code, but there are various parts of the code like this line where I intentionally copy the array larger than the original to zero-extend a new array to fill in afterwards. Some byte array idioms :)

PR #92, opened 2021-09-04
Avoid inefficient usage of arraylist

FastAtlas:
Hi,
We find that there are three ArrayList objects which are not manipulated by random access. Due to the memory reallocation triggered in the successive insertions, the time complexity of add method of ArrayList is amortized O(1). We notice that these objects are only used for traversal and the retrieval for the first or the last element.

This functionality can be implemented by LinkedList. Moreover, the insertion of LinkedList is strictly O(1) time complexity because no memory reallocation occurs.

We discovered this inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/bm/Lang.java
+++ b/src/main/java/org/apache/commons/codec/language/bm/Lang.java
@@ -18,6 +18,7 @@
 package org.apache.commons.codec.language.bm;
 
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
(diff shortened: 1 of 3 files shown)


DittoTool:
Besides, we also found several TreeMap objects, which are only used for key-value pair retrieval. It would be better to replace it with HashMap.

jochenw:
Are there any sources, that you can quote with regards to the complexity of ArrayList vs. LinkedList? Thanks, Jochen

jochenw:
Besides: What is the tool Ditto, that you ae referring to?

garydgregory:
I do not think we should change internals until we have an example that demonstrates whatever problem this tool thinks it found. This feels backward to me because accessing the first and/last element of an array is fast and an array list grows in chunks.

garydgregory:
Tree maps are also handy for reproducing the same result over and over, also needs a demonstration of what the problem is in pragmatic terms.

jochenw:
> I do not think we should change internals until we have an example that demonstrates whatever problem this tool thinks it found. This feels backward to me because accessing the first and/last element of an array is fast and an array list grows in chunks.

I think exactly the same, which is why was asking for more infos.

garydgregory:
Closing, no reply from OP.

PR #94, opened 2021-09-22
CODEC-303: Refactor NoOpBaseNCodec to improve test logic

wx930910:
### Jira

- [CODEC-303](https://issues.apache.org/jira/browse/CODEC-303)

#### Replace test class [NoOpBaseNCodec](https://github.com/apache/commons-codec/blob/b70e177f8733e9dd4e5f503620ed0b136b74785b/src/test/java/org/apache/commons/codec/binary/BaseNCodecTest.java#L375) by mocking object and improve test design
<hr>

##### Motivation

 * Decouple test class `NoOpBaseNCodec` from production class `BaseNCodec`.
 * Remove the redundant overridden methods.
 * Make testing logic more explict.

<hr>

##### Key changed/added classes in this PR
 * Created mocking object to replace test subclass `NoOpBaseNCodec`, decoupled test from production code.
 * Remove test subclass `NoOpBaseNCodec`.
 * Created method that return the mocking object for code reuse.
 * Add Mockito dependency.

<hr>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -243,6 +243,12 @@ limitations under the License.
       <version>3.12.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>3.12.4</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
--- a/src/test/java/org/apache/commons/codec/binary/BaseNCodecTest.java
+++ b/src/test/java/org/apache/commons/codec/binary/BaseNCodecTest.java
@@ -28,9 +28,19 @@ import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 
+import static org.mockito.Mockito.CALLS_REAL_METHODS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.withSettings;
+
 public class BaseNCodecTest {
 
-    BaseNCodec codec;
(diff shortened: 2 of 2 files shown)


wx930910:
We can also get rid of the method that return the mocking object and directly create the mocking object inside of each test case.

garydgregory:
While this seems like a clever use of Mockito, it hinders ease of understanding the code and maintenance IMO.

Mocking is a great technique for certain use cases, this isn't one of them IMO. The current noop adapter is clear, simple, and can be stepped into while debugging if needes.

wx930910:
@garydgregory, thanks for providing the feedback! What do you think about [this case](https://github.com/apache/commons-jci/blob/4006b2c36ccd5aac05f844f4b86895370afa36af/fam/src/test/java/org/apache/commons/jci2/monitor/FilesystemAlterationMonitorTestCase.java#L128), do you think we will get benefit from Mockito by replacing the inheritance with spying object? IMO It seems like the only reason we use inheritance is because we cannot directly create an instance from abstract class.

garydgregory:
Using Mockito here I feel falls under the old adage "Once you have a hammer, everything starts to look like a nail." The class in question is an abstract class and therefore cannot be instantiated, the proper use case is subclassing, both for an app and therefore for a test. Using Mockito to avoid using a concrete subclass only obfuscates what the test is trying to do, makes the code harder to maintain and to learn as you are now requiring new users and maintainers to learn Mockito on top of learning the code base.
In other code bases, a noop concrete type might be part of the main tree instead of in the test tree and this whole thread would likely not even come up.
IOW, the shortcut Mockito affords is not worth the complexity it requires to learn and maintain.

wx930910:
@garydgregory I see, thanks you much for providing such detailed feedback!

PR #104, opened 2022-02-13
Maven Wrapper v3.1.0 with Maven v3.8.4

nhojpatrick:
Now the Takari Maven Wrapper has been adopted by Apache Maven, I'm submitted again.
This makes cicd simpler so now the cicd process can be used to check the latest maven works.
Each project does need to be individually managed but also if projects require a specific maven version, it can be pinned.


garydgregory:
My preference is to not clutter each component with this noise, it's redundant.

nhojpatrick:
@garydgregory i agree it could be considered clutter. If all projects are kept active current it's never an issue.
From experience I'm coming from working with dead or hibernated projects, when moving company/job/team/project and having to kick start something building. It use to work on the cicd server, but someone updated that to a newer maven version or java version so the project i'm working doesn't work anymore. So 1st things I now always setup is maven wrapper (takari before it was merge) and enforcer, so the project itself knows was it was last configured to build under.

garydgregory:
I'm still -1 for now. We have GitHub Actions builds, we used to also have Jenkins builds, but I'm not sure we need those unless we want to push SNAPSHOTS from within Apache. Forcing whatever pile of scripts and files in the root of the project makes a bigger mess, and what makes it worse and is confusing is then looking at whatever files are used to define CI builds. Fortunately, I recently dropped most Travis CI config files, so that leaves us with GitHub and Jenkins (which is defined elsewhere)

nhojpatrick:
WIll close my maven wrapper pr's for apache commons. Giving up trying to advocate the benefits.

PR #107, opened 2022-02-19
Inline variable.

arturobernalg:


src/main/java/org/apache/commons/codec/digest/UnixCrypt.java
```diff
@@ -326,8 +326,7 @@ public static String crypt(final String original, final String salt) {
     }
 
     private static int byteToUnsigned(final byte b) {
-        final int value = b;
-        return value < 0 ? value + 256 : value;
+        return (int) b < 0 ? (int) b + 256 : (int) b;
```
aherbert, line 329: return b & 0xff;

arturobernalg: Changed.

PR #109, opened 2022-02-19
CODEC-285 HexTest

nhojpatrick:
Looking at @Test(expected) and test method body, the last line usually throws the expected but test actually fails earlier.

### DIFF
--- a/src/test/java/org/apache/commons/codec/binary/HexTest.java
+++ b/src/test/java/org/apache/commons/codec/binary/HexTest.java
@@ -556,20 +556,18 @@ public class HexTest {
         assertArrayEquals("64".toCharArray(), hex);
     }
 
-    @Test(expected=ArrayIndexOutOfBoundsException.class)
+    @Test
     public void testEncodeHexPartialInputUnderbounds() {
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/codec/binary/HexTest.java
```diff
@@ -556,20 +556,18 @@ public void testEncodeHexPartialInput() {
...
+        assertThrows(ArrayIndexOutOfBoundsException.class, () -> Hex.encodeHex(data, -2, 10, true));
     }
 
-    @Test(expected=ArrayIndexOutOfBoundsException.class)
+    @Test
     public void testEncodeHexPartialInputOverbounds() {
         final byte data[] = "hello world".getBytes(StandardCharsets.UTF_8);
 
-        final char[] hex = Hex.encodeHex(data, 9, 10, true);
-        assertArrayEquals("64".toCharArray(), hex);
```
kinow, line 572: Ah, the `assertArrayEquals` was never called. Good catch :+1: Thanks.

kinow (approved):

PR #110, opened 2022-02-19
CODEC-285 remove unused imports

nhojpatrick:


kinow (approved):
Thanks for the contribution @nhojpatrick ! For small changes like this, especially if changing imports & docs in test files, it's alright to skip the JIRA issue. If your modifications are more substantial (e.g. upgrade Junit version, re-write test base classes) then a JIRA issue might be helpful in case a dev finds an issue working with the tests later. But this PR and #109 and 110 look quite simple & small :+1:

PR #116, opened 2022-03-18
Update README.md

Dmole:
The central Maven repository only has 1.15 (not 1.16)
https://mvnrepository.com/artifact/commons-codec/commons-codec


garydgregory:
-1: This is the master branch for work in progress toward the next release, not the release tag.

Dmole:
@garydgregory there is no README.md on the 1.5 branch, maybe the pom example should be removed from the README.md because it will always be wrong?

Dmole:
eg:

https://github.com/apache/commons-codec/pull/117

PR #126, opened 2022-04-16
Avoid use toString() or substring()in favor of a simplified expression.

arturobernalg:


src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java
```diff
@@ -403,7 +403,7 @@ public String encode(String input, final Languages.LanguageSet languageSet) {
         input = input.toLowerCase(Locale.ENGLISH).replace('-', ' ').trim();
 
         if (this.nameType == NameType.GENERIC) {
-            if (input.length() >= 2 && input.substring(0, 2).equals("d'")) { // check for d'
+            if (input.length() >= 2 && input.startsWith("d'")) { // check for d'
```
garydgregory, line 406: @arturobernalg 
Why still check for length >=2 ? What is change automatically generated through IntelliJ or Eclipse?

PR #135, opened 2022-06-12
CODEC-305: Fix byte-skipping in Base16 decoding

gpsfl:


kinow (commented):
Thanks for your contribution! I pressed the button to allow the workflow to run, but didn't review the code @gpsfl. Looks like the build failed. Could you take a look, maybe try running `mvn` locally to confirm the build passes on your environment first, if you don't want to wait for GH Actions to run.

gpsfl:
Thank you for the quick response. It looks like builds on the master branch were already failing before my changes with the same error (see [here](https://github.com/apache/commons-codec/commit/89cb9572ccdff84082c37e14c49d4ac6cb32f4f5) for example). I was also able to reproduce it locally, both with and without my changes. Should I perhaps use another branch for this?

kinow:
> Thank you for the quick response. It looks like builds on the master branch were already failing before my changes with the same error (see [here](https://github.com/apache/commons-codec/commit/89cb9572ccdff84082c37e14c49d4ac6cb32f4f5) for example). I was also able to reproduce it locally, both with and without my changes. Should I perhaps use another branch for the pull request?

Ah, thanks for looking into that @gpsfl . In that case we should fix the build and rebase this one to confirm these changes work as well. Feel free to wait until someone else steps in and fixes the build (or if you have spare time and feel like fixing that, pull requests are always welcome, though sometimes it can be tricky to figure out how to fix the build).

Bruno

src/main/java/org/apache/commons/codec/binary/Base16.java
```diff
@@ -177,26 +177,24 @@ void decode(final byte[] data, int offset, final int length, final Context conte
...
 
             // reset to empty-value for next invocation!
             context.ibitWorkArea = 0;
         }
 
-        while (i < charsToProcess) {
-            result = decodeOctet(data[offset++]) << BITS_PER_ENCODED_BYTE;
-            result |= decodeOctet(data[offset++]);
-            i += 2;
+        while (i + 1 < charsToProcess) {
```
aherbert, line 188: I would prefer to not use `i + 1 < charsToProcess` on the loop condition.

It seems the logic setting `charsToProcess` is incorrect when there is a previous half-byte. The value should not be final and can be decremented by 1 if `i` is set to 1:
```
result |= decodeOctet(data[offset++]);
i = 1;
charsToProcess--;

// ...

while (i < charsToProcess) {
```

This corrected logic still requires that offset is added to i: `data[offset + i++]`. Previously offset was incremented.

This may be better served by computing the end point for offset:
```Java
// End is an even number of chars added to the current offset
final int end = offset + (dataLen & ~0x1);
while (offset < end) {
            result = decodeOctet(data[offset++]) << BITS_PER_ENCODED_BYTE;
            result |= decodeOctet(data[offset++]);
}

if (offset < dataLen) {

```

This logic would eliminate the requirement for `i` and `charsToProcess`. The logic to handle the ensureBufferSize can just use `availableChars`. Note: The code is untested.

aherbert (changes requested):
Thanks for the fix. It seems the method logic is a bit convoluted. I have suggested a change that will consume the trailing 1/2 byte if present, then consume an even number of bytes from the remaining input length, then save a trailing 1/2 byte if present.

Can you review this and see if it passes your new unit test. Thanks.

gpsfl:
I have implemented your suggestions and ran all unit tests successfully (using IntelliJ because the maven build is still broken).

aherbert:
Looks good. Can you rebase on master as I have fixed the build. Thanks.

kinow:
> Looks good. Can you rebase on master as I have fixed the build. Thanks.

I think @gpsfl found out that the build on `master` is broken - https://github.com/apache/commons-codec/pull/135#issuecomment-1153063366

If so we may need to fix that so that this PR can be rebased and hopefully pass the build.

aherbert:
@kinow I just fixed the master build. It was due to animal-sniffer plugin complaining about a breaking API change in JDK 9+ with Buffer -> ByteBuffer as a return object from ByteBuffer operations. So a rebase should pick this up.

kinow:
> @kinow I just fixed the master build. It was due to animal-sniffer plugin complaining about a breaking API change in JDK 9+ with Buffer -> ByteBuffer as a return object from ByteBuffer operations. So a rebase should pick this up.

Thanks @aherbert !!!

gpsfl:
I have rebased the changes from master

aherbert:
Looks fine. Can you squash all changes and prefix the commit message with `CODEC-305: Fix byte-skipping in Base16 decoding`. Thanks.

gpsfl:
> Looks fine. Can you squash all changes and prefix the commit message with `CODEC-305: Fix byte-skipping in Base16 decoding`. Thanks.

Done

gpsfl:
Can you give any ETA on 1.16 release?

aherbert:
A 1.16 RC1 was trialled in January. However the release process was halted by an upstream dependency. I shall raise a question on the dev mailing list to see if the release manager has an update on this.

gpsfl:
> A 1.16 RC1 was trialled in January. However the release process was halted by an upstream dependency. I shall raise a question on the dev mailing list to see if the release manager has an update on this.

Any news regarding this?

aherbert:
The issue was related to upstream dependencies for managing the release. From the conversation on the mailing list:
```
"The current issue is that if a component does not use the
latest Apache RAT plugins, then it must do so in its POM. This could be
handled automatically if we released a new commons-parent."
```

PR #136, opened 2022-06-29
Remove redundant operations.

arturobernalg:
Remove redundant operations. Clear operations on empty collections have no effect and can be removed.

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/DaitchMokotoffSoundex.java
+++ b/src/main/java/org/apache/commons/codec/language/DaitchMokotoffSoundex.java
@@ -494,9 +494,6 @@ public class DaitchMokotoffSoundex implements StringEncoder {
 
             for (final Rule rule : rules) {
                 if (rule.matches(inputContext)) {
-                    if (branching) {
-                        nextBranches.clear();
-                    }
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/codec/language/DaitchMokotoffSoundex.java
```diff
@@ -494,9 +494,6 @@ private String[] soundex(final String source, final boolean branching) {
 
             for (final Rule rule : rules) {
                 if (rule.matches(inputContext)) {
-                    if (branching) {
-                        nextBranches.clear();
-                    }
```
kinow, line 499: Isn't that to force a `clear()` when `nextBranches` is not an empty collection?

```
nextBranches = branching ? new ArrayList<>() : Collections.emptyList();
```

arturobernalg: 'nextBranches' is always empty collections.  The loop will be exit  in the line 525  with the `break` in the first iterations or/and when the rule match ` if (rule.matches(inputContext))`. So, at the end its always empty and its redundant.

PR #139, opened 2022-07-15
(chore) Fix minor typos

marcwrobel:
Fix various minor typos in :

- project documentations,
- javadoc and comments,
- method attribute names,
- test method names.


src/main/java/org/apache/commons/codec/binary/Base32.java
```diff
@@ -124,11 +124,6 @@ public class Base32 extends BaseNCodec {
     // The private member fields below are used with the new streaming approach, which requires
     // some state be preserved between calls of encode() and decode().
 
-    /**
-     * Place holder for the bytes we're dealing with for our based logic.
-     * Bitwise operations store and extract the encoding or decoding from this variable.
-     */
```
kinow, line 130: Note to self: check out in git history if there was a variable below this comment before, or if this comment was intended for a block of variables?

kinow (approved):
Looks good to me! Some are grammar or writing style changes (e.g. case-sensitive vs. case sensitive), but no objections from me :+1: 

Thank you @marcwrobel !

src/main/java/org/apache/commons/codec/binary/Base32.java
kinow: This comment was for `bitWorkArea` — https://github.com/apache/commons-codec/commit/d6f305d2efd7560634217a25f3bbee59299844dd

That field was removed in CODEC-96, so safe to remove this comment now :detective:  — https://github.com/apache/commons-codec/commit/3f41cf1609fae7c8848beb0655bcf67132c7a277

kinow:
Merged, thanks @marcwrobel !

PR #189, opened 2023-06-26
CODEC-308: change NYSIIS encoding to not remove the first character i…

Ben-Waters:
With the current implementation of NYSIIS, it is possible to incorrectly remove the first character from the encoding. 

According to the algorithm the first character of the string should be the first character of the encoding, then based on a bunch of other rules are applied to the string characters are removed. The implementation in commons-codec passes the entire string into the transcodeRemaining method which works for the most part and then afterwards, checks that there is at least 1 character before removing the final 'A' or 'S'. 

The problem is, if you have a word like "ASH" you will end up with a single final character of "A". Similarly with "SSH" you would have "S" and the logic will currently remove it and return a blank string when it should still return at least the first letter of the original string.

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/Nysiis.java
+++ b/src/main/java/org/apache/commons/codec/language/Nysiis.java
@@ -275,7 +275,8 @@ public class Nysiis implements StringEncoder {
 
         // First character of key = first character of name.
         final StringBuilder key = new StringBuilder(str.length());
-        key.append(str.charAt(0));
+        final char firstChar = str.charAt(0);
+        key.append(firstChar);
 
         // Transcode remaining characters, incrementing by one character each time
         final char[] chars = str.toCharArray();
@@ -313,6 +314,12 @@ public class Nysiis implements StringEncoder {
             // If last character is A, remove it.
             if (lastChar == 'A') {
                 key.deleteCharAt(key.length() - 1);
+            }
+
+            if(key.length()==0){
+                // We've removed the first character of the string. Likely because it was an S or A
+                // We should return at least the first character
+                key.append(firstChar);
             }
         }
 
--- a/src/main/java/org/apache/commons/codec/language/Nysiis.java
+++ b/src/main/java/org/apache/commons/codec/language/Nysiis.java
@@ -316,7 +316,7 @@ public class Nysiis implements StringEncoder {
                 key.deleteCharAt(key.length() - 1);
             }
 
-            if(key.length()==0){
+            if (key.length() == 0) {
                 // We've removed the first character of the string. Likely because it was an S or A
                 // We should return at least the first character
                 key.append(firstChar);
--- a/src/main/java/org/apache/commons/codec/language/Nysiis.java
+++ b/src/main/java/org/apache/commons/codec/language/Nysiis.java
@@ -308,7 +308,7 @@ public class Nysiis implements StringEncoder {
                 key.deleteCharAt(key.length() - 1);
             }
 
-            if(key.length()==0){
+            if (key.length() == 0) {
                 // We've removed the first character of the string. Likely because it was an S or A
                 // We should return at least the first character
                 key.append(firstChar);
--- a/src/main/java/org/apache/commons/codec/language/Nysiis.java
+++ b/src/main/java/org/apache/commons/codec/language/Nysiis.java
@@ -267,7 +267,8 @@ public class Nysiis implements StringEncoder {
 
         // First character of key = first character of name.
         final StringBuilder key = new StringBuilder(str.length());
-        key.append(str.charAt(0));
+        final char firstChar = str.charAt(0);
+        key.append(firstChar);
 
         // Transcode remaining characters, incrementing by one character each time
         final char[] chars = str.toCharArray();
@@ -305,6 +306,12 @@ public class Nysiis implements StringEncoder {
             // If last character is A, remove it.
             if (lastChar == 'A') {
                 key.deleteCharAt(key.length() - 1);
+            }
+
+            if(key.length()==0){
+                // We've removed the first character of the string. Likely because it was an S or A
+                // We should return at least the first character
+                key.append(firstChar);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/codec/language/NysiisTest.java
```diff
@@ -140,7 +140,8 @@ public void testDropBy() throws EncoderException {
                 new String[] { "JILES", "JAL" },
                 // violates 6: if the last two characters are AY, remove A
                 new String[] { "CARRAWAY", "CARY" },       // Original: CARAY
-                new String[] { "YAMADA", "YANAD" });
+                new String[] { "YAMADA", "YANAD" },
+                new String[] { "ASH", "A"});
```
garydgregory, line 144: Hi @Ben-Waters 
Thank you for your PR.
Who is right? This or Fuzzy?
```
py
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import fuzzy
>>> fuzzy.nysiis('ASH')
'AS'
>>>
```
Thoughts?

Ben-Waters: Based on this [other implementation](http://www.dropby.com/indexLF.html?content=/NYSIIS.html) it would be just A.

According to the [algorithm](https://en.wikipedia.org/wiki/New_York_State_Identification_and_Intelligence_System), the final character should be removed if it is an 'S' so I would think it should be removed.

The current commons-codec implementation is removing it as well as the final 'A' but is ignoring the part about "The first character of the NYSIIS code is the first character of the name."

Ben-Waters: I haven't used fuzzy before but looking at [their implementation](https://github.com/yougov/fuzzy/blob/master/src/fuzzy.pyx#L81) I don't see where they remove the trailing S or A.

garydgregory: I had not used it before today either. I was looking for _something_ easy to compare this PR's behavior. 

Let's see if anybody else has feedback. 

Do you happen to know of anything else that is easily testable for this use-case?

Ben-Waters: It's pretty frustrating but it seems like the results are inconsistent depending on where you go.

Looking at [this link](https://rosettacode.org/wiki/NYSIIS) the Java version returns AS but the python returns A. For my use-case, either one of those would be fine but it seems hard to get a consistent answer. Commons-codec is currently returning nothing though and I haven't seen another do that yet.

I don't know of a good official implementation to run it against.

garydgregory: I would think that returning something like "A" is better than "". Let's see if anyone else suggests anything. @kinow ?

garydgregory:
@Ben-Waters Not directly related, but do you have any thoughts on #281?

src/test/java/org/apache/commons/codec/language/NysiisTest.java
garydgregory: The question that comes up for me is: Is our current implementation "plain" NYSIIS, or, is any non-standard or any behavior from the "modified" L&A 1977 algorithm present in our current implementation? If our current code is "plain", then I could see us creating a `ModifiedNysiis` class, if not, then we are in a bit of a pickle.

kinow: Although the results seem to indicate we implement the plain NYSIIS, someone has indeed to verify it reading our code and the algorithm. A good point, +1!

Ben-Waters:
> @Ben-Waters Not directly related, but do you have any thoughts on #36?



> @Ben-Waters Not directly related, but do you have any thoughts on #36?

Hmmm I'm no expert on this since it isn't NYSIIS but some other algorithm but I can take a look.

src/test/java/org/apache/commons/codec/language/NysiisTest.java
Ben-Waters: Based on the [steps from wikipedia](https://en.wikipedia.org/wiki/New_York_State_Identification_and_Intelligence_System#cite_ref-taft_2-0), the core of the issue is that in step 4, we are setting the pointer to the first character instead of the 2nd character. The 9th step on wikipedia is a bit ambiguous though as to whether that step includes the first character or not. If it does, then you would get a blank string like commons-codec currently gives. If not, then you would get 'A'.

Unfortunately I don't have access to the original source book for the algorithm to try to clarify. It looks like [berkeley law](https://lawcat.berkeley.edu/record/38122) has it but I don't have an account for that.

garydgregory:
I just re-read the comments it seems like:
- We are not sure if the current code implements the "plain" or original algorithm.
- We don't have access to, or cannot find, the paper for the original algorithm.
- If we implement the plain original algorithm, then we can add a new class for the newer "modifed" algorithm.
- If we do not implement the plain original algorithm, then we need to talk about that.

Help needed.

src/main/java/org/apache/commons/codec/language/Nysiis.java
```diff
@@ -314,6 +315,12 @@ public String nysiis(String str) {
             if (lastChar == 'A') {
                 key.deleteCharAt(key.length() - 1);
             }
+
+            if (key.length() == 0) {
+                // We've removed the first character of the string. Likely because it was an S or A
+                // We should return at least the first character
+                key.append(firstChar);
+            }
```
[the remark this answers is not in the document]
Ben-Waters: Copilot is wrong on this. The check to remove a solitary 's' is already happening before this final length check, at lines 300-303

src/test/java/org/apache/commons/codec/language/NysiisTest.java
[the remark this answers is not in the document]
Ben-Waters: I've added in a new test case, "SSH" which becomes "S". I ran it locally and it did give the expected result.

PR #208, opened 2023-09-28
CODEC-310 - Documentation update for the org.apache.commons.codec.digest.* package

yakovsh:
This PR:
* Documents that SecureRandom is used in Crypt, Md5Crypt and Sha2Crypt classes by changing Javadocs in these classes (by calling B64.randomSalt)
* Changes salt generation in UnixCrypt to use SecureRandom to match the other classes
* Update the userguide to list all functions from the digest package
* Changes the hyperlinks in the user guide from HTTP to HTTPS

See https://issues.apache.org/jira/browse/CODEC-310


garydgregory:
Patch applied. TY @yakovsh!

PR #219, opened 2023-11-22
CODEC-311: Fix possible ArrayIndexOutOfBoundException thrown by RefinedSoundex.getMappingCode()

arthurscchan:
This fixes a possible ArrayIndexOutOfBoundException in [src/main/java/org/apache/commons/codec/language/RefinedSoundex.java](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/language/RefinedSoundex.java) thrown by `RefinedSoundex.getMappingCode()` method when handling out of range letters.

This PR adds a conditional checking to ensure the index is never out of bounds from the configured soundexMapping array. If the calculated index goes out of bounds, it will simply return 0, just like the original logic when Character.isLetter() returns false.

We found this bug using fuzzing by way of OSS-Fuzz. It is reported at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64353.

### DIFF
--- a/src/test/java/org/apache/commons/codec/language/RefinedSoundexTest.java
+++ b/src/test/java/org/apache/commons/codec/language/RefinedSoundexTest.java
@@ -80,6 +80,16 @@ public class RefinedSoundexTest extends AbstractStringEncoderTest<RefinedSoundex
     }
 
     @Test
+    public void testInvalidSoundexCharacter() {
+        final char[] invalid = new char[256];
+        for (int i = 0; i < invalid.length; i++) {
+            invalid[i] = (char)i;
+        }
+
+        assertEquals(new RefinedSoundex().encode(new String(invalid)), "A0136024043780159360205050136024043780159360205053");
+    }
+
+    @Test
     public void testNewInstance() {
         assertEquals("D6043", new RefinedSoundex().soundex("dogs"));
     }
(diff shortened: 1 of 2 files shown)


garydgregory:
Hello @arthurscchan 
You'll need a unit test to prove this change works.

arthurscchan:
@garydgregory Thanks for the comment. I have added a unit test.

src/test/java/org/apache/commons/codec/language/RefinedSoundexTest.java
```diff
@@ -93,4 +94,13 @@ public void testNewInstance2() {
     public void testNewInstance3() {
         assertEquals("D6043", new RefinedSoundex(RefinedSoundex.US_ENGLISH_MAPPING_STRING).soundex("dogs"));
     }
+
+    @Test
+    public void testInvalidSoundexCharacter() {
+        char[] invalid = new char[256];
+        for (int i = 0; i < 256; i++) {
```
garydgregory, line 101: Hello @arthurscchan 
You'll want to remove use of the magic number here and say `invalid.length`.
TY!

garydgregory (changes requested):

src/test/java/org/apache/commons/codec/language/RefinedSoundexTest.java
arthurscchan: Have changed. Thanks.

garydgregory:
@arthurscchan 

PR #221, opened 2023-11-22
CODEC-313: Fix possible ArrayIndexOutOfBoundsException thrown by QuotedPrintableCodec.encodeQuotedPrintable() method

arthurscchan:
This fixes a possible ArrayIndexOutOfBoundException in [src/main/java/org/apache/commons/codec/language/QuotedPrintableCodec.java](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/language/QuotedPrintableCodec.java) thrown by `QuotedPrintableCodec.encodeQuotedPrintable()` method when the input byte array has less than 3 elements.

This PR adds a conditional check to ensure the index is never negative. It will simply return null if the byte array is too short (with a length less than 3) if strict value is true.

We found this bug using fuzzing by way of OSS-Fuzz. It is reported at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64358.

### DIFF
--- a/src/main/java/org/apache/commons/codec/net/QuotedPrintableCodec.java
+++ b/src/main/java/org/apache/commons/codec/net/QuotedPrintableCodec.java
@@ -85,6 +85,11 @@ public class QuotedPrintableCodec implements BinaryEncoder, BinaryDecoder, Strin
     private static final byte LF = 10;
 
     /**
+     * Minimum length required for the byte arrays used by encodeQuotedPrintable method
+     */
+    private static final int MIN_BYTES = 3;
+
+    /**
      * Safe line length for quoted printable encoded text.
      */
     private static final int SAFE_LENGTH = 73;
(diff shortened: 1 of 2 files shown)


garydgregory:
@arthurscchan 
You'll need a unit test to prove what the main change does.

arthurscchan:
Hi, I have added the unit test.

garydgregory:
@arthurscchan 

src/main/java/org/apache/commons/codec/net/QuotedPrintableCodec.java
```diff
@@ -208,6 +208,11 @@ public static final byte[] encodeQuotedPrintable(BitSet printable, final byte[]
         final int bytesLength = bytes.length;
 
         if (strict) {
+            // Ensure bytes have at least 3 bytes
+            if (bytesLength < 3) {
```
garydgregory, line 212: Hello @arthurscchan 
If the number `3` is special, please refactor it into a constant and document it.

garydgregory (changes requested):
@arthurscchan
Please see my comment.

garydgregory:
@arthurscchan 
TY, merged. It looks like encoder exceptions are not in play here.

PR #222, opened 2023-11-22
CODEC-314: Fix possible IndexOutOfBoundsException thrown by PercentCodec.insertAlwaysEncodeChars() method

arthurscchan:
This fixes a possible IndexOutOfBoundsException in [src/main/java/org/apache/commons/codec/net/PercentCodec.java](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/net/PercentCodec.java) thrown by `PercentCodec.insertAlwaysEncodeChars()` method if the initial byte array contains negative bytes which are used as index for the BitSet.

This PR adds a conditional check to ensure only valid bytes (positive or zero) are processed.

We found this bug using fuzzing by way of OSS-Fuzz. It is reported at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64362.

### DIFF
--- a/src/main/java/org/apache/commons/codec/net/PercentCodec.java
+++ b/src/main/java/org/apache/commons/codec/net/PercentCodec.java
@@ -77,10 +77,15 @@ public class PercentCodec implements BinaryEncoder, BinaryDecoder {
      *
      * @param alwaysEncodeChars the unsafe characters that should always be encoded
      * @param plusForSpace      the flag defining if the space character should be encoded as '+'
+     * @throws EncoderException if the alwaysEncodeChars byte array contains invalid bytes
      */
-    public PercentCodec(final byte[] alwaysEncodeChars, final boolean plusForSpace) {
+    public PercentCodec(final byte[] alwaysEncodeChars, final boolean plusForSpace) throws EncoderException {
         this.plusForSpace = plusForSpace;
-        insertAlwaysEncodeChars(alwaysEncodeChars);
+        try {
+            insertAlwaysEncodeChars(alwaysEncodeChars);
+        } catch (IndexOutOfBoundsException e) {
+            throw new EncoderException(e);
+        }
     }
 
     private boolean canEncode(final byte c) {
--- a/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
+++ b/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
@@ -106,6 +106,13 @@ public class PercentCodecTest {
     }
 
     @Test
+    public void testInvalidByte() throws Exception {
+        final byte[] invalid = {(byte)-1, (byte)'A'};
+
+        assertThrows(EncoderException.class, () -> new PercentCodec(invalid, true));
+    }
+
+    @Test
     public void testPercentEncoderDecoderWithNullOrEmptyInput() throws Exception {
         final PercentCodec percentCodec = new PercentCodec(null, true);
         assertNull(percentCodec.encode(null), "Null input value encoding test");


garydgregory:
Let's just have ONE PR instead of one-liner PRs. Especially since these all claim to fix the exact same kind of issue. Also, you MUST provide a matching test.

arthurscchan:
Hi, I have added unit test. I am not sure about combining the PRs since they are opened as 4 different issues in the JIRA.

garydgregory (changes requested):
Hello @arthurscchan 
Please see my comment.

garydgregory:
> Hi, I have added unit test. I am not sure about combining the PRs since they are opened as 4 different issues in the JIRA.

In this case, I don't care about mapping PRs to JIRA 1-1. YMMV ;-) I care about making my life easier ;-)

garydgregory:
Hi @arthurscchan,

Please help me understand why we need this PR because at first glance this PR could be making the use case worse IMO:
- Before PR: Grabage in -> exception, albeit not `IllegalArgumentException`.
- After PR: Garbage in -> garbage out

Or am I missing something?
TY!

arthurscchan:
Hi @garydgregory,

In my understanding, the constructor of PecentCodec calls the `insertAlwaysEncodeChars(byte[])` method in [Line 83](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/net/PercentCodec.java#L83). The use of the `insertAlwaysEncodeChars(byte[])` method is to record all bytes in the provided byte array to be some character that requires encode all the time. Thus if garbage bytes do exist in the byte array, it should be skipped and all other "valid" bytes should still be considered. For example, if someone thinks that the letter A should always be encoded, and accidentally provides `byte[] encode = {(byte)'A', (byte) -1};`, the logic could still work, skipping the -1 bytes and change the 'A' byte in the BitSet to true. Also, the byte array itself is never stored in the object. Thus the PR could make the running smoother without crashing even if some invalid byte does exist. The new logic simply skips them and continues. The most important part is that "valid" bytes are still processed and interpreted.
This is my understanding, am I missing something here??

arthurscchan:
> > Hi, I have added unit test. I am not sure about combining the PRs since they are opened as 4 different issues in the JIRA.
> 
> In this case, I don't care about mapping PRs to JIRA 1-1. YMMV ;-) I care about making my life easier ;-)

If you want, I don't mind combining all the PRs together, not sure if it is more messy to handle though.

garydgregory:
@arthurscchan 

garydgregory:
> @arthurscchan Please use a better description in PRs and JIRA: Specify the class and method where the exception occurs.



> > > Hi, I have added unit test. I am not sure about combining the PRs since they are opened as 4 different issues in the JIRA.
> > 
> > 
> > In this case, I don't care about mapping PRs to JIRA 1-1. YMMV ;-) I care about making my life easier ;-)
> 
> If you want, I don't mind combining all the PRs together, not sure if it is more messy to handle. Although they are throwing similar exceptions, I would say the reason is not too similar IMO.

Well, at this point, I commented on all PRs, so we might as well leave it as is. 
You'll have to make sure to review all comments from all PRs as they apply to all PRs but I did not want to repeat myself all over the place ;-)

garydgregory:
> Hi @garydgregory,
> 
> In my understanding, the constructor of PecentCodec calls the `insertAlwaysEncodeChars(byte[])` method in [Line 83](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/net/PercentCodec.java#L83). The use of the `insertAlwaysEncodeChars(byte[])` method is to record all bytes in the provided byte array to be some character that requires encoding. Thus if garbage bytes do exist in the byte array, it should be skipped and all other "valid" bytes should still be considered. For example, if someone thinks that the letter A should always be encoded, and accidentally provides `byte[] encode = {(byte)'A', (byte) -1};`, the logic could still work, skipping the -1 bytes and change the 'A' byte in the BitSet to true. Also, the byte array itself is never stored in the object. Thus the PR could make the running smoother without crashing even if some invalid byte does exist. The new logic simply skips them and continues. The most important part is that "valid" bytes are still processed and interpreted. This is my understanding, am I missing something here?? Do let me know if I interpret the functionality wrongly. Thanks.

Hmm, so that means that one can roundtrip encode/decode, but when you start with bad data, you end up with nothing (if all the input is bad) or good encoded data. But when you decode that encoded data, you don't end up with what you started. If we really want all of that, I think we should update the Javadoc and describe that encoding also performs filtering and can never throw an exception for illegal input. WDYT?
Does that mean encode never throws EncoderException?

arthurscchan:
@garydgregory Oh, you are right, if we filtered out invalid data, then the round trip result will be differed from the original input. Sorry for missing that. Maybe throwing EncoderException is much better. Instead of checking the length and skip the invalid byte, I could simply add a try catch block in the code, when any IndexOutOfBound (or more general, and kind of RuntimeException) is thrown, wrap it with EncoderException. This could at least make the method not throwing "unexpected" RuntimeException. It maybe a better approach. WDYT?

arthurscchan:
Hi @garydgregory,

I think it is a different case. The invalid bytes are provided through constructor, it is only used for the initial setting of the PercentCodec object to set which bytes are to be encoded during the encode or decode method.  Thus if we initialise the PercentCodec with a byte array contains invalid bytes, it just mean that the encode method will not encode that byte. It does not affect the encode and decode process at all. Thus the encode / decode method round trip won't be affected and the result is the same as the original input. Maybe I could add a round trip call and ensure the result are the same in the test case. 
Alternatively, instead of checking the length and skip the invalid byte during constructor phase, I could simply add a try catch block in the code, when any IndexOutOfBound (or more general, and kind of RuntimeException) is thrown, wrap it with EncoderException. This could at least make the method not throwing "unexpected" RuntimeException. It maybe a better approach. WDYT?

arthurscchan:
Hi @garydgregory, OK. Sorry for not considering in that direction. I will change the code in this PR to wrap around the possible IndexOutOfBoundException in an EncoderException, to at least decrease the change of "unpexected" input thrown directly from the constructor. That maybe a better approach. Thanks for your comments.

garydgregory:
TY @arthurscchan 
Let's keep in mind that this is a library that needs to make sense (or try to) for users. As opposed to just addressing fuzzing issues in the most expedient manner ;-) TY again for your work :-)

src/main/java/org/apache/commons/codec/net/PercentCodec.java
```diff
@@ -77,10 +77,15 @@ public PercentCodec() {
      *
      * @param alwaysEncodeChars the unsafe characters that should always be encoded
      * @param plusForSpace      the flag defining if the space character should be encoded as '+'
+     * @throws EncoderException if the alwaysEncodeChars byte array contains invalid bytes
      */
-    public PercentCodec(final byte[] alwaysEncodeChars, final boolean plusForSpace) {
+    public PercentCodec(final byte[] alwaysEncodeChars, final boolean plusForSpace) throws EncoderException {
```
garydgregory, line 82: Hi @arthurscchan 
Now that I look at this again, it doesn't make sense to me, this seems like the exact use case for an `IllegalArgumentException`. In addition, the constructor does not encode anything, so an `EncoderException` really does not make sense IMO.

I propose:
```
diff --git a/src/main/java/org/apache/commons/codec/net/PercentCodec.java b/src/main/java/org/apache/commons/codec/net/PercentCodec.java
index 8e51538..df623df 100644
--- a/src/main/java/org/apache/commons/codec/net/PercentCodec.java
+++ b/src/main/java/org/apache/commons/codec/net/PercentCodec.java
@@ -231,6 +231,9 @@
      * @param b the byte that is candidate for min and max limit
      */
     private void insertAlwaysEncodeChar(final byte b) {
+        if (b < 0) {
+            throw new IllegalArgumentException("byte must be >= 0");
+        }
         this.alwaysEncodeChars.set(b);
         if (b < alwaysEncodeCharsMin) {
             alwaysEncodeCharsMin = b;
diff --git a/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java b/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
index 34f8ecb..a537efb 100644
--- a/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
+++ b/src/test/java/org/apache/commons/codec/net/PercentCodecTest.java
@@ -106,6 +106,12 @@
     }
 
     @Test
+    public void testInvalidByte() throws Exception {
+        final byte[] invalid = { (byte) -1, (byte) 'A' };
+        assertThrows(IllegalArgumentException.class, () -> new PercentCodec(invalid, true));
+    }
+
+    @Test
     public void testPercentEncoderDecoderWithNullOrEmptyInput() throws Exception {
         final PercentCodec percentCodec = new PercentCodec(null, true);
         assertNull(percentCodec.encode(null), "Null input value encoding test");
```
WDYT?

garydgregory (changes requested):
Hi @arthurscchan 
Rethought this a bit, please see my comment.

src/main/java/org/apache/commons/codec/net/PercentCodec.java
arthurscchan: Yes, I think it is a better changes and interpretation of the exceptions. I will change it accordingly.
@garydgregory

arthurscchan:
@garydgregory Sorry I am bit busy and did not handle that last week. Thanks for fixing that for me.

PR #223, opened 2023-11-23
CODEC-315: Fix possible IndexOutOfBoundException thrown by PhoneticEngine.encode method

arthurscchan:
This fixes possible StringIndexOutOfBoundsException and ArrayIndexOutOfBoundsException in [src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java](https://github.com/apache/commons-codec/blob/master/src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java) thrown by `PhoneticEngine.encode()` when the provided string only contains one of the name prefix of the chosen NameType or only contain single quotation character.

This PR fixes the parameter for the split method and adds a conditional check to ensure only strings and arrays are not empty before processing.

We found this bug using fuzzing by way of OSS-Fuzz. It is reported at https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64376 and https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=64395.

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java
+++ b/src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java
@@ -409,7 +409,7 @@ public class PhoneticEngine {
         switch (this.nameType) {
         case SEPHARDIC:
             words.forEach(aWord -> {
-                final String[] parts = aWord.split("'");
+                final String[] parts = aWord.split("'", -1);
                 words2.add(parts[parts.length - 1]);
             });
             words2.removeAll(NAME_PREFIXES.get(this.nameType));
@@ -431,7 +431,7 @@ public class PhoneticEngine {
         } else if (words2.size() == 1) {
             // not a multi-word name
             input = words.iterator().next();
-        } else {
+        } else if (!words2.isEmpty()) {
             // encode each word in a multi-word name separately (normally used for approx matches)
             final StringBuilder result = new StringBuilder();
             words2.forEach(word -> result.append("-").append(encode(word)));
--- a/src/test/java/org/apache/commons/codec/language/bm/PhoneticEngineTest.java
+++ b/src/test/java/org/apache/commons/codec/language/bm/PhoneticEngineTest.java
@@ -48,6 +48,15 @@ public class PhoneticEngineTest {
                 );
     }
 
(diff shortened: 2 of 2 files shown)


src/test/java/org/apache/commons/codec/language/bm/PhoneticEngineTest.java
```diff
@@ -70,4 +80,13 @@ public void testEncode(final String name, final String phoneticExpected, final N
...
         }
     }
+
+    @ParameterizedTest
+    @MethodSource("invalidData")
+    public void testInvalidEncode(final String input, final NameType nameType, final RuleType ruleType,
+                                  final boolean concat, final int maxPhonemes) {
+        final PhoneticEngine engine = new PhoneticEngine(nameType, ruleType, concat, maxPhonemes);
+
+        assertDoesNotThrow(() -> engine.encode(input));
```
garydgregory, line 90: Why is `assertDoesNotThrow` used here? The test method does nothing else after the call to encode so there is nothing else to run. Or, am I missing something?

src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java
```diff
@@ -431,7 +431,7 @@ public String encode(String input, final Languages.LanguageSet languageSet) {
         } else if (words2.size() == 1) {
             // not a multi-word name
             input = words.iterator().next();
-        } else {
+        } else if (words2.size() > 0) {
```
garydgregory, line 434: Use `isEmpty()`.

garydgregory:
@arthurscchan 

src/test/java/org/apache/commons/codec/language/bm/PhoneticEngineTest.java
arthurscchan: Originally I just want to test if the invalid input still throw any unexpected exceptions. But yes, I do agree the result also needed to be checked. I have changed the test case a little bit, as well as for other PRs. Thanks for your suggestions.

src/main/java/org/apache/commons/codec/language/bm/PhoneticEngine.java
arthurscchan: Changed. Thanks for your suggestion.

PR #229, opened 2024-01-08
modifying data size in DigestUtilsTest to speed up MessageDigestAlgor…

TestImprove:
The runtime of the test class `MessageDigestAlgorithmsTest` can be improved by modifying the data array size in `DigestUtilsTest`.

The byte array returned by `getTestData()` comes from DigestUtilsTest and has a size of `1024 * 1024`. When this array size is reduced to `32 * 32`, for example, the runtime of the `MessageDigestAlgorithmsTest` test class drops from `7.046 s` to `1.032 s` on our local machine. 

We propose not to make the tests in `MessageDigestAlgorithmsTest.java` rely on the data created by `DigestUtilsTest.java`.

### DIFF
--- a/src/test/java/org/apache/commons/codec/digest/DigestUtilsTest.java
+++ b/src/test/java/org/apache/commons/codec/digest/DigestUtilsTest.java
@@ -56,7 +56,7 @@ public class DigestUtilsTest {
 
     private static final String EMPTY_STRING = "";
 
-    private final byte[] testData = new byte[1024 * 1024];
+    private final byte[] testData = new byte[32 * 32];
 
     private Path testFile;
 
--- a/src/test/java/org/apache/commons/codec/digest/DigestUtilsTest.java
+++ b/src/test/java/org/apache/commons/codec/digest/DigestUtilsTest.java
@@ -56,7 +56,7 @@ public class DigestUtilsTest {
 
     private static final String EMPTY_STRING = "";
 
-    private final byte[] testData = new byte[32 * 32];
+    private final byte[] testData = new byte[1024 * 4];
 
     private Path testFile;
 


garydgregory:
-1: The test data variable contains "randomized" data, and taking your argument further, we could make the array size 1, which would reduce even more the randomness of the data. Of course, using randomized data can be argued back and forth but that's an additional way to test.

aherbert:
Note: The `DigestTestUtils` and `MessageDigestAlgorithmsTest` classes are creating a random `byte[]` and writing it to two files for every method. If this behaviour is changed from `Before/AfterEach` to `Before/AfterAll` (with use of static data) then I observe an speed-up of 20%.

Since the data is non-destructively used, then a simple switch to make the `testFile` and `testRandomAccessFile` the same saves 10% by requiring 1 less file write per test fixture.

Given the limited impact of optimising the test resources, the major runtime of the test is the actual digest of the megabyte of random data.

The internal buffer size in `DigestUtils` is 1024. So if the test data is changed then it should be larger than this size. The suggested 32*32 (1024) is too small to require looping within `DigestUtils` when digesting a `FileChannel` or `InputStream`.

I would suggest updating the tests to use a single source of random data created once. The size can be reduced to improve speed but it should not compromise testing of the digest algorithms.

garydgregory:
@aherbert 
That sounds good.

TestImprove:
Thanks for the explanation. Since you pointed out that `32 * 32` is too small, we wonder what do you think about changing the byte array size to `1024 * 2 or 3` instead of `1024 * 1024`? just want to know the reason behind `1024 * 1024`.

Also, we are curious why some tests in `MessageDigestAlgorithmsTest` have the same statement twice. For example in the `testDigestByteArray()` test method, the statement `assertArrayEquals(digestTestData(messageDigestAlgorithm), DigestUtils.digest(DigestUtils.getDigest(messageDigestAlgorithm), getTestData()));` is written twice.

garydgregory:
> Thanks for the explanation! Since you pointed out that `32 * 32` is too small, we wonder what do you think about changing the byte array size to `1024 * 2 or 3` instead of `1024 * 1024`? just want to know the reason behind `1024 * 1024`.

@TestImprove 
I think you're asking a question that was already answered above by @aherbert

> 
> Also, we are curious why some tests in `MessageDigestAlgorithmsTest` have the same statement twice. For example in the `testDigestByteArray()` test method, the statement `assertArrayEquals(digestTestData(messageDigestAlgorithm), DigestUtils.digest(DigestUtils.getDigest(messageDigestAlgorithm), getTestData()));` is written twice.

Without taking the time to analyze this example, I can say there are many ways to test code, one of which is testing for idempotency and side effects.

aherbert:
There would not be an issue changing from `1024 * 1024` to `n * 1024` where `n > 1`. Try setting it at 4 and testing if the speed is satisfactory.

Note that some test functions will duplicate calls on the same input to check that the input is not destructively modified.

garydgregory:
I'm going to close this as I don't see the gain compared to the benefit of running with more random samples because:

- Note that without any change the whole CI build takes less than 2 minutes. 
- On my PC, running this test class from Eclipse takes 0.7 seconds.
- From the command line, the test takes 1.2 seconds: 
```
[INFO] Tests run: 36, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.203 s -- in org.apache.commons.codec.digest.DigestUtilsTest
```

PR #240, opened 2024-02-05
[CODEC-263] test for user reported issue

elharo:
Near as I can tell this issue doesn't reproduce and can be closed. @garydgregory

### DIFF
--- a/src/test/java/org/apache/commons/codec/binary/Base64Test.java
+++ b/src/test/java/org/apache/commons/codec/binary/Base64Test.java
@@ -1309,6 +1309,10 @@ public class Base64Test {
 
     }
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/codec/binary/Base64Test.java
```diff
@@ -1309,6 +1309,11 @@ public void testUrlSafe() {
 
     }
 
+    public void testCodec263(){
+      Base64.decodeBase64("publishMessage");
```
garydgregory, line 1313: I would make it more obvious IMO to refactor the magic strings into a local variable (or a constant).

PR #262, opened 2024-04-03
only initialize digestUtilsTest when necessary to speed up test MessageDigestAlgorithmsTest

TestBoost:
Sorry but I don't have a jira account yet and I just submit this pull request firstly through Github now.

For the test methods in the test class MessageDigestAlgorithmsTest, not all test methods modify the contents of shared fields. We can just initialize `digestUtilsTest` once before all the test methods to speed up this test. However, the parameterized tests for test method `testNonBlockingDigestRandomAccessFile` always change the the file-pointer offset of the `RandomAccessFile` field in `digestUtilsTest` class.

There are two ways to fix this problem: one is to reinitialize `digestUtilsTest` before the test method `testNonBlockingDigestRandomAccessFile` like we propose in this pull request; another is to add `randomAccessFile.seek(0);`
 at the end of test method `testNonBlockingDigestRandomAccessFile`.

The test runtime can jump from `5.509 s` to `5.046 s` after applying the changes when run on our machine.

### DIFF
--- a/src/test/java/org/apache/commons/codec/digest/MessageDigestAlgorithmsTest.java
+++ b/src/test/java/org/apache/commons/codec/digest/MessageDigestAlgorithmsTest.java
@@ -36,6 +36,7 @@ import java.nio.file.StandardOpenOption;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
+import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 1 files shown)


garydgregory:
-1: This is not an issue worth making the test more convoluted, especially with the odd code pattern of calling tear down and set up methods manually.

garydgregory:
Closing; see my previous comment.

PR #300, opened 2024-07-29
Update Sha2Crypt.java

kangxy:
The old regular expression is incorrect.
If you want to match [a-zA-Z0-9./] and limit the length to 16, I think it should be the following expression： "^\\$([56])\\$(rounds=(\\d+)\\$)?([\\.\\/a-zA-Z0-9]{1,16})$")

### DIFF
--- a/src/main/java/org/apache/commons/codec/digest/Sha2Crypt.java
+++ b/src/main/java/org/apache/commons/codec/digest/Sha2Crypt.java
@@ -69,7 +69,7 @@ public class Sha2Crypt {
 
     /** The pattern to match valid salt values. */
     private static final Pattern SALT_PATTERN = Pattern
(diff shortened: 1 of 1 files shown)


garydgregory:
@kangxy 

There are no tests in this PR.
You'll want a test that fails if there are no changes to the main directory.

kangxy:
> The proposed RE does not work because it does not allow for truncating the salt after 16 characters, as per the following test failures:
> 
> Sha256CryptTest.testSha256CryptStrings:63 » IllegalArgument Invalid salt value: $5$1234567890123456789 Sha512CryptTest.testSha512CryptStrings:98 » IllegalArgument Invalid salt value: $6$1234567890123456789
> 
> There needs to be further work before the RE can be made stricter.

Both of the above salt values are wrong. We need the salt to be 16 bytes or less

PR #347, opened 2024-12-11
Simplify hex methods

Jatie1:
Update methods encodeHex and decodeHex to simplify the code and make it slightly more efficient.

encodeHex: don't compute dataOffset + dataLen on every for iteration and also standardize nibble high and low using Base16.encode() as inspiration. I would say it's easier to read now.

decodeHex: get high and low nibble directly using for loop iterator incrementing by 2 (2 char per hex val) and OR them immediately instead of getting high nibble, iterating variable, ORing low nibble, iterating variable then converting to byte. I think it's far more readable this way.

I tested both with a barebones jmh benchmark and speed is roughly the same or very slightly faster than old implementation for both.


garydgregory:
Hello @Jatie1 

Please update the PR so anyone can run your JMH benchmarks.

Thank you!

garydgregory:
Closing: No reply in 2 weeks.

PR #353, opened 2024-12-27
Use Resource.class to load resources, rather than its classloader

msfroh:
This plays nicer with Java's module system, as the class is able to load files from within the current module, whereas the classloader may not necessarily be able to.

Thanks for your contribution to [Apache Commons](https://commons.apache.org/)! Your help is appreciated!

Before you push a pull request, review this list:

- [x] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [x] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

### DIFF
--- a/src/main/java/org/apache/commons/codec/Resources.java
+++ b/src/main/java/org/apache/commons/codec/Resources.java
@@ -33,7 +33,7 @@ public class Resources {
      * @return An input stream.
      */
     public static InputStream getInputStream(final String name) {
-        final InputStream inputStream = Resources.class.getClassLoader().getResourceAsStream(name);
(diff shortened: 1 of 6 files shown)


garydgregory:
https://issues.apache.org/jira/browse/CODEC-324

garydgregory:
TY @msfroh !

PR #438, opened 2026-06-18
[CODEC-344] Validate Base64 custom alphabets

OldTruckDriver:
[CODEC-344] Validate Base64 custom alphabets

Base64.Builder#setEncodeTable(byte...) accepted invalid custom alphabets. Duplicate entries could create a codec that failed to decode its own output, and alphabets containing the configured padding byte could make encoded data be interpreted as padding.

This change validates custom Base64 encode tables, rejects duplicate entries, rejects tables containing the configured padding byte, and uses unsigned byte lookup so non-ASCII alphabet bytes can round-trip consistently.

Tests cover duplicate entries, configured padding-byte conflicts, non-ASCII alphabet bytes, and the case where '=' is valid after changing the padding byte.

Tests run:
- mvn -q -Dtest=org.apache.commons.codec.binary.Base64Test test
- mvn -q

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -45,6 +45,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
     <release version="1.22.1" date="YYYY-MM-DD" description="This is a feature and maintenance release. Java 8 or later is required.">
       <!-- FIX -->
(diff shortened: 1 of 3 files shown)


garydgregory:
Jira ticket https://issues.apache.org/jira/browse/CODEC-344
PR merged 🚀 Thank you @OldTruckDriver

PR #440, opened 2026-06-29
Compile language encoder regexes once (Caverphone1/2, MatchRating)

nishantmehta:
## Summary

Three phonetic encoders in the `language` package applied their regular expressions via `String.replaceAll`, which **recompiles its regex argument on every call**. A single `encode` therefore compiled many `Pattern`s, repeated for every input.

This hoists those regexes into `static final Pattern` constants (compiled once) and applies them with `Matcher.replaceAll`. The regexes, their order, and the replacements are unchanged, so the produced codes are identical.

- **Caverphone2** — ~25 patterns per encode
- **Caverphone1** — 17 patterns per encode
- **MatchRatingApproachEncoder** — 9 patterns per encode (`cleanName`'s five punctuation regexes + a whitespace-collapse applied three times + a boundary whitespace-run in `removeVowels`)

## Measurement

ThreadMXBean allocation driver, 200k warmed ops:

| encoder | before | after |
|---------|--------|-------|
| `Caverphone2.encode` | 19670 B/op | 6304 B/op (−68%) |
| `Caverphone1.encode` | 14005 B/op | 4672 B/op (−67%) |
| `MatchRatingApproachEncoder.encode` | ~5800 B/op | ~1800 B/op (−70%) |

## Testing

`Caverphone1Test`, `Caverphone2Test` and `MatchRatingApproachEncoderTest` pass unchanged.

Three commits, one per encoder, in case you prefer to take them separately.

### DIFF
--- a/src/main/java/org/apache/commons/codec/language/Caverphone1.java
+++ b/src/main/java/org/apache/commons/codec/language/Caverphone1.java
@@ -18,6 +18,7 @@
 package org.apache.commons.codec.language;
 
 import java.util.Locale;
+import java.util.regex.Pattern;
 
 /**
  * Encodes a string into a Caverphone 1.0 value.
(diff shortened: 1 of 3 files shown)


garydgregory:
Closing. No one's asking for this. Fix bugs in Jira if want to help improve this things.

PR #444, opened 2026-09-15
Make the @param names in BaseNCodec match the parameters they document

hxperl:
Three `@param` tags in `BaseNCodec` name a parameter their method does not have. javadoc does not render a `@param` it cannot bind, so rather than documenting the argument, each of these caused that argument's documentation to be dropped.

Reproduced on `master` at `f49b96388c2d` with nothing but the JDK:

```
$ javadoc -Xdoclint:reference,syntax -private -quiet -d /tmp/jd \
      org/apache/commons/codec/binary/BaseNCodec.java
org/apache/commons/codec/binary/BaseNCodec.java:483: error: @param name not found
```

Renamed the three tags to match their parameters. The errors are gone and no new ones appear.


garydgregory:
Thank you @hxperl , merged 🚀

